Issue #5, opened 2016-08-22
Issues running on Windows


brettcannon:
E.g. on Windows the binary is named `python.exe` -- same on macOS -- and the binary is stored in a `Scripts` directory instead of `bin` (#2 tries to solve this). `os.execv()` also doesn't quite operate the way you think on (at least) Windows 10 as it immediately returns to PowerShell/cmd.exe and the process runs in the background of the prompt, seeming to pause on I/O unless the user interacts with the terminal, e.g. hitting enter  (if you type `exit` at the prompt then the new process takes over the terminal window, although if you kill *that* process then you kill the entire terminal window instead of returning, including the end of the benchmark run).

haypo:
For os.execv(): it's a quick hack to enhance the usability of the bench.py script, it avoids to have to enter manually the virtual environment.

Do you know if https://virtualenvwrapper.readthedocs.io/ or https://pypi.python.org/pypi/vex works on Windows? If yes, we should check how they work.

brettcannon:
I don't know if they work but I can give them a try. Otherwise we should consider exposing a function in `venv` in the stdlib to figure this sort of thing out.

ncoghlan:
I like vex for this kind of problem, as it keeps things simple: it just starts a subshell with the venv activated.

The main subprocess launching is in https://github.com/sashahart/vex/blob/master/vex/run.py

For invocation "vex <benchvenv-name> perf ...." should do the trick.

haypo:
The commit 4f74a9df1948c6fa2f04acb69ab0ceaf4a48472a fixed the most obvious issue, fix path to the Python program inside the virtual environment.

haypo:
You should be able to workaround the issue by creating manually the virtual environment and then using the internal --inside-venv option. Something like:

```
virtualenv env
env/bin/python -m pip install "setuptools>=18.5" "pip>=6.0"
env/bin/python -m pip install -r performance/requirements.txt 
env/bin/python -m performance run --inside-venv (...)
```

ammaraskar:
>os.execv() also doesn't quite operate the way you think on (at least) Windows 10 as it immediately returns to PowerShell/cmd.exe and the process runs in the background of the prompt, seeming to pause on I/O unless the user interacts with the terminal

This was getting fairly annoying for me so I've opened up #8 as a quick fix for it until you guys evaluate using vex/venvwrapper

haypo:
[...]

PR #8, opened 2016-08-25
Change execv to Popen to avoid cmd backgrounding issue in windows.

ammaraskar:
This is the issue specifically: https://bugs.python.org/issue19124
This is also mentioned by Brett in #5

### DIFF
--- a/performance/venv.py
+++ b/performance/venv.py
@@ -240,4 +240,12 @@ def create_virtualenv(python):
 def exec_in_virtualenv(options):
     venv_python = create_virtualenv(options.python)
     args = [venv_python, "-m", "performance"] + sys.argv[1:] + ["--inside-venv"]
-    os.execv(args[0], args)
+    # os.execv() is buggy on windows, which is why we use run_cmd/subprocess
+    # on windows. 
+    # * https://bugs.python.org/issue19124
+    # * https://github.com/python/benchmarks/issues/5
+    if os.name == "nt":
+        run_cmd(args)
+        sys.exit(0)
+    else:
+        os.execv(args[0], args)


performance/venv.py
```diff
@@ -240,4 +240,4 @@ def create_virtualenv(python):
 def exec_in_virtualenv(options):
     venv_python = create_virtualenv(options.python)
     args = [venv_python, "-m", "performance"] + sys.argv[1:] + ["--inside-venv"]
-    os.execv(args[0], args)
+    run_cmd(args)
```
haypo: Hum. Can you please keep os.execv() for UNIX/BSD, but use run_cmd() for other platforms?

And please also add sys.exit(0) after run_cmd().

ammaraskar: >Hum. Can you please keep os.execv() for UNIX/BSD, but use run_cmd() for other platforms?

Any particular reason why?

>And please also add sys.exit(0) after run_cmd().

That's what the return in the `cli.py` file does. I think the caller should do it explicitly.

haypo: > Any particular reason why?

It avoids to have two running processes. Using os.execv(), you only get one process. It might help to get more reliable benchmarks.

> That's what the return in the cli.py file does. I think the caller should do it explicitly.

Using os.execv(), the function never returns. So it's better to have the same behaviour if os.execv() or run_cmd() is used.

ammaraskar: >It avoids to have two running processes. Using os.execv(), you only get one process. It might help to get more reliable benchmarks.

Seems kinda contrived but may as well be cautious as far as benchmarking goes.

haypo:
Cool, thanks for the fix. It will help to find the next Windows bug :-)

PR #9, opened 2016-08-30
Update Django to 1.10

haypo:


haypo:
I don't know well Django so I would prefer to get a review on this change.

cc @brettcannon @alex @fijal.

alex:
Keep in mind, this is changing the code that's running, so it's really a different benchmark.

haypo:
alex: "Keep in mind, this is changing the code that's running, so it's really a different benchmark."

I know. I consider that you should only compare benchmark results produced with the same version of the benchmark suite. At this point, I don't really care of backward compatibility. I prefer to upgrade dependencies to be more representative of nowadays applications.

The previous Python benchmark suite used a different approach of duplicating completely benchmarks to upgrade a library. Because it's more complex to upgrade libraries, old versions of libraries were used. For example, it uses Mercurial 1.2.1 which was released in 2009 (7 years ago).

But you are right that we need a clear policy about upgrading library. For example, use semantic versionning and increase the major version number when a library is upgraded to notice that results become incompatible?

haypo:
FYI I increased the "major" version from 0.1.x to 0.2.0 because Django was upgraded.

PR #10, opened 2016-09-01
add fractions benchmarks

scoder:
Add fractions benchmarks that compare the decimal and fractions modules using the same benchmarking code. 
See https://bugs.python.org/issue22458

### DIFF
--- a/performance/benchmarks/__init__.py
+++ b/performance/benchmarks/__init__.py
@@ -227,15 +227,9 @@ def BM_Telco(python, options):
 
 
 @VersionRange()
-def BM_Telco_Fractions(python, options):
-    bm_path = Relative("bm_telco_fractions.py")
+def BM_Fractions(python, options):
+    bm_path = Relative("bm_fractions.py")
     return run_perf_script(python, options, bm_path, extra_args=['fractions'])
-
-
-@VersionRange()
-def BM_Telco_Decimal(python, options):
-    bm_path = Relative("bm_telco_fractions.py")
-    return run_perf_script(python, options, bm_path, extra_args=['decimal'])
 
 
 @VersionRange()
--- a/performance/benchmarks/__init__.py
+++ b/performance/benchmarks/__init__.py
@@ -227,6 +227,18 @@ def BM_Telco(python, options):
 
 
 @VersionRange()
+def BM_Telco_Fractions(python, options):
+    bm_path = Relative("bm_telco_fractions.py")
+    return run_perf_script(python, options, bm_path, extra_args=['fractions'])
+
+
+@VersionRange()
+def BM_Telco_Decimal(python, options):
+    bm_path = Relative("bm_telco_fractions.py")
+    return run_perf_script(python, options, bm_path, extra_args=['decimal'])
+
+
+@VersionRange()
 def BM_Hexiom2(python, options):
     bm_path = Relative("bm_hexiom2.py")
     return run_perf_script(python, options, bm_path)
--- a/performance/benchmarks/bm_telco_fractions.py
+++ b/performance/benchmarks/bm_telco_fractions.py
@@ -67,9 +67,8 @@ def run(cls, loops=1):
     return perf.perf_counter() - start
 
 
-def run_bench(n, backend_class):
-    run(backend_class, loops=2)  # warmup
-    return run(backend_class, n)
+def run_bench(loops, backend_class):
+    return run(backend_class, loops)
 
 
 def find_benchmark_class(impl_name):
--- a/performance/benchmarks/bm_telco_fractions.py
+++ b/performance/benchmarks/bm_telco_fractions.py
@@ -16,6 +16,8 @@ The total price, t, is converted to a string, s.
 
 """
 
+from __future__ import absolute_import
+
 from struct import unpack
 import os.path
 
--- a/performance/benchmarks/bm_telco_fractions.py
+++ b/performance/benchmarks/bm_telco_fractions.py
@@ -67,11 +67,11 @@ def run(cls, loops=1):
 
             sumT += t
 
-        results.add(str(sumT))
+            results.add('%.4f' % t)
 
     time = perf.perf_counter() - start
 
-    assert len(results) == 1
+    assert 1 <= len(results) <= len(values)
     return time
 
 
--- a/performance/benchmarks/bm_telco_fractions.py
+++ b/performance/benchmarks/bm_telco_fractions.py
@@ -43,6 +43,7 @@ def run(cls, loops=1):
             values.append(n)
 
     start = perf.perf_counter()
+    results = set()
     for _ in xrange(loops):
         sumT = cls()   # sum of total prices
         sumB = cls()   # sum of basic tax
@@ -64,7 +65,12 @@ def run(cls, loops=1):
 
             sumT += t
 
-    return perf.perf_counter() - start
+        results.add(str(sumT))
+
+    time = perf.perf_counter() - start
+
+    assert len(results) == 1
+    return time
 
 
 def run_bench(loops, backend_class):
--- a/performance/benchmarks/bm_telco_fractions.py
+++ b/performance/benchmarks/bm_telco_fractions.py
@@ -20,6 +20,7 @@ from struct import unpack
 import os.path
 
 import perf
+from six.moves import xrange
 
 
 def rel_path(*path):
@@ -35,15 +36,14 @@ def run(cls, loops=1):
 
     values = []
     with open(filename, "rb") as infil:
-        for _ in range(20000):
+        for _ in xrange(20000):
             datum = infil.read(8)
             if datum == '': break
             n, =  unpack('>Q', datum)
             values.append(n)
 
     start = perf.perf_counter()
-
-    for _ in range(loops):
+    for _ in xrange(loops):
         sumT = cls()   # sum of total prices
         sumB = cls()   # sum of basic tax
         sumD = cls()   # sum of 'distance' tax
--- a/performance/benchmarks/bm_telco_fractions.py
+++ b/performance/benchmarks/bm_telco_fractions.py
@@ -2,8 +2,8 @@
 
 """ Telco Benchmark for measuring the performance of Fraction calculations
 
-http://www2.hursley.ibm.com/decimal/telco.html
-http://www2.hursley.ibm.com/decimal/telcoSpec.html
+http://speleotrove.com/decimal/telco.html
+http://speleotrove.com/decimal/telcoSpec.html
 
 A call type indicator, c, is set from the bottom (least significant) bit of the duration (hence c is 0 or 1).
 A r, r, is determined from the call type. Those calls with c=0 have a low r: 0.0013; the remainder (‘distance calls’) have a ‘premium’ r: 0.00894. (The rates are, very roughly, in Euros or dollarates per second.)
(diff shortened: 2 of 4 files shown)


haypo:
> Add fractions benchmarks that compare the decimal and fractions modules using the same benchmarking code.

I'm not sure that I understand the purpose of the change. Do you want to compare the performances of the fractions module with the performance of the decimal module?

It's not really how the "performance" module is used. This module is a set of benchmarks to compare the performance of two Python implementations.

Maybe we can add the benchmark, use fractions by default, but don't run the benchmark with decimal? I mean that you should run it with decimal manually.

performance/benchmarks/__init__.py
```diff
@@ -227,6 +227,18 @@ def BM_Telco(python, options):
...
 @VersionRange()
+def BM_Telco_Fractions(python, options):
+    bm_path = Relative("bm_telco_fractions.py")
+    return run_perf_script(python, options, bm_path, extra_args=['fractions'])
+
+
+@VersionRange()
+def BM_Telco_Decimal(python, options):
+    bm_path = Relative("bm_telco_fractions.py")
+    return run_perf_script(python, options, bm_path, extra_args=['decimal'])
```
haypo: As I wrote in the comment, I don't think that it makes sense to test the decimal module?

performance/benchmarks/bm_telco_fractions.py
```diff
@@ -0,0 +1,100 @@
+#-*- coding: UTF-8 -*-
+
+""" Telco Benchmark for measuring the performance of Fraction calculations
+
+http://www2.hursley.ibm.com/decimal/telco.html
```
haypo: HTTP 404 error :-/

performance/benchmarks/bm_telco_fractions.py
```diff
@@ -0,0 +1,100 @@
...
+http://www2.hursley.ibm.com/decimal/telcoSpec.html
```
haypo: HTTP 404 error :-/

Hum, in a different flavor of the bm_telco.py benchmark, I found this link. Is it accurate?
http://speleotrove.com/decimal/telco.html

performance/benchmarks/bm_telco_fractions.py
```diff
@@ -0,0 +1,100 @@
...
+
+A call type indicator, c, is set from the bottom (least significant) bit of the duration (hence c is 0 or 1).
+A r, r, is determined from the call type. Those calls with c=0 have a low r: 0.0013; the remainder (‘distance calls’) have a ‘premium’ r: 0.00894. (The rates are, very roughly, in Euros or dollarates per second.)
+A price, p, for the call is then calculated (p=r*n).
+A basic tax, b, is calculated: b=p*0.0675 (6.75%), and the total basic tax variable is then incremented (sumB=sumB+b).
+For distance calls: a distance tax, d, is calculated: d=p*0.0341 (3.41%), and then the total distance tax variable is incremented (sumD=sumD+d).
+The total price, t, is calculated (t=p+b, and, if a distance call, t=t+d).
+The total prices variable is incremented (sumT=sumT+t).
+The total price, t, is converted to a string, s.
```
haypo: Is that true? :-) I don't see the conversion to string. Maybe update the docstring?

performance/benchmarks/bm_telco_fractions.py
```diff
@@ -0,0 +1,100 @@
...
+from struct import unpack
+import os.path
+
+import perf
+
+
+def rel_path(*path):
+    return os.path.join(os.path.dirname(__file__), *path)
+
+filename = rel_path("data", "telco-bench.b")
```
haypo: I suggest:

DATA_FILE = os.path.join(os.path.dirname(__file__), "data", "telco-bench.b")

Oh, I see that this code is also used in the current bm_telco.py :-) So maybe keep it. I might change it later ;-)

performance/benchmarks/bm_telco_fractions.py
```diff
@@ -0,0 +1,100 @@
...
+                sumD += d
+                t += d
+
+            sumT += t
+
+    return perf.perf_counter() - start
+
+
+def run_bench(n, impl):
+    if impl == 'fractions':
```
haypo: run_bench() is called many times, once per sample (by default, perf uses at least 4 samples including 1 warmup sample). Maybe put this code in the "main" body?

scoder:
What I mean is that there is one benchmark implementation that is executed with two different backends, thus giving *comparable* results for two different stdlib libraries. Thus, I think it's good to execute both. The results are not directly comparable with the "bm_telco" benchmark because that does some unrelated processing along the way.

performance/benchmarks/bm_telco_fractions.py
```diff
@@ -0,0 +1,100 @@
...
+                d = p * disttax
+                sumD += d
+                t += d
+
+            sumT += t
+
+    return perf.perf_counter() - start
+
+
+def run_bench(n, impl):
```
haypo: I suggest to rename "n" to "loops" to be consistent with other benchmarks.

performance/benchmarks/bm_telco_fractions.py
```diff
@@ -0,0 +1,100 @@
...
+    if impl == 'fractions':
+        from fractions import Fraction as backend_class
+    elif impl == 'quicktions':
+        from quicktions import Fraction as backend_class
+    elif impl == 'decimal':
+        from decimal import Decimal as backend_class
+    else:
+        raise ValueError("Invalid class name: '%s'" % impl)
+
+    run(backend_class, loops=2)  # warmup
```
haypo: Don't do that! run_bench() itself is used to compute warmup samples...

Moreover, perf automatically runs a first warmup sample by default (and more warmup samples on PyPy). The perf module does that for you.

Try to run your script using "--worker --verbose" options (or just "--verbose") to see all computes samples.

performance/benchmarks/bm_telco_fractions.py
```diff
@@ -0,0 +1,100 @@
...
+    return run(backend_class, n)
+
+
+def prepare_subprocess_args(runner, args):
+    args.append(runner.args.backend)
+
+
+if __name__ == "__main__":
+    import perf.text_runner
+    runner = perf.text_runner.TextRunner(name='telco')
```
haypo: Error: bm_telco.py already uses this name. Rename it to telco_fractions?

performance/benchmarks/bm_telco_fractions.py
```diff
@@ -0,0 +1,100 @@
...
+    runner.metadata['description'] = "Telco fractions benchmark"
+    runner.prepare_subprocess_args = prepare_subprocess_args
+
+    parser = runner.argparser
+    backends = ["decimal", "fractions", "quicktions"]
+    parser.add_argument("backend", choices=backends, nargs='?', default="fractions")
+
+    options = runner.parse_args()
+    runner.name += "/%s" % options.backend
```
haypo: Hum, you cannot use the "python3 -m perf compare" command to compare decimal with fractions if the two benchmarks have a different name. I suggest to not modify the benchmark name, but only run the benchmark with the fractions module for the benchmark suite.

performance/benchmarks/bm_telco_fractions.py
scoder: Everything in this benchmark is rather fast, but sure, I can move it out.

performance/benchmarks/bm_telco_fractions.py
scoder: The goal is not to get a benchmark run that compares both, but to get comparable numbers for both. These benchmarks are quite fast, why not keep both?

performance/benchmarks/bm_telco_fractions.py
scoder: If you look a little further down, you will see that this is not actually the final name. The problem is just that the runner is needed for parsing the arguments. Chicken and egg.

scoder:
I've updated the pull request.

performance/benchmarks/bm_telco_fractions.py
skrah: Indeed.  Just "Based on the telco benchmark".

performance/benchmarks/__init__.py
skrah: Agreed, unless you want to point out the speed difference.  But that was my question on bugs.python.org: What is the specific goal of this benchmark? :)

performance/benchmarks/bm_telco_fractions.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def run_bench(loops, backend_class):
+    return run(backend_class, loops)
+
+
+def find_benchmark_class(impl_name):
+    if impl_name == 'fractions':
+        from fractions import Fraction as backend_class
+    elif impl_name == 'quicktions':
```
skrah: Hmm, what about gmpy2 rationals?  As I understand it, this test suite is about regressions within Python itself. I'm not sure this would be a good precedent.

performance/benchmarks/bm_telco_fractions.py
```diff
@@ -0,0 +1,110 @@
...
+            t = p + b
+
+            if calltype:
+                d = p * disttax
+                sumD += d
+                t += d
+
+            sumT += t
+
+        results.add(str(sumT))
```
skrah: I think this should be in the loop, and should use (for fractions) `print("%s" % (Decimal(t.numerator)/Decimal(t.denominator)))`

performance/benchmarks/bm_telco_fractions.py
skrah: I'd just use `bm_fractions.py`.

performance/benchmarks/bm_telco_fractions.py
scoder: It's not a decimal specific benchmark, though. It's one that can be used with decimal. Thus, formatting should depend on the library that is being benchmarked and not a specific other one.

skrah: Yes, definitely. Just the output should be as expected for a financial calculation.

scoder: I can change it to "%.4f" formatting, would that work for you? I think that's a reasonable thing to do with a fraction after computing it.

performance/benchmarks/bm_telco_fractions.py
scoder: I can't see a reason why the benchmarks shouldn't support non-stdlib libraries that reimplement standard Python functionality, as long as it can be achieved with a reasonable amount of adaptation. The elementtree benchmark I've written also supports lxml (AFAICT, it's been modified since), you can pass the library to import by name.

And since you asked, yes, gmpy2 would certainly qualify, but rather for the original telco benchmark, right?

scoder:
I've updated the pull request to address the comments.

haypo:
I closed the issue https://bugs.python.org/issue22458 but please take a look at the discussion there.

I closed the CPython issue to suggest to continue the discussion on this (GitHub) bug tracker.

haypo:
Sorry, but I'm not really convinced that a fractions benchmark is really helpful to compare the performances of different Python implementations.

If you still want a fractions benchmark, maybe you can send me a pull request to my https://github.com/haypo/pymicrobench project which is a much wider collection of random CPython (micro)benchmarks. This project has a different purpose.

At least, I took your update on the Telco URL :-D I also wrote a longer description for each benchmark in the documentation. Here is the new documentation for telco:
http://pyperformance.readthedocs.io/benchmarks.html#telco

I close the PR.

PR #14, opened 2016-10-07
Add MDP benchmark

haypo:
Code comes from PyPy benchmarks:
https://bitbucket.org/pypy/benchmarks/src/59290b59a24e54057d4c694fa4f47e7879a347a0/own/bm_mdp.py?at=default&fileviewer=file-view-default

Changes:
- Reformat using autopep8
- Port to Python 3: replace zip()[0] with list(zip())[0]
- Remove dead code like tarjanSCC()
- Use the perf module to run the benchmark

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -304,6 +304,7 @@ Available Benchmarks
     ignored
 
 - ``mako`` - use the Mako template system to build a 150x150-cell HTML table.
+- ``mdp`` - battle with damages and topological sorting of nodes in a graph
 - ``meteor_contest`` - solver for Meteor Puzzle board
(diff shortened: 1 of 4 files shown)


haypo:
I had a small issue with the benchmark: a sample takes 8 seconds (CPython 2.7, CPython 3.7 or PyPy take between 7 and 8 seconds). So a full run is... long. I don't know how to tune to the benchmark to reduce the duration of a single sample.

haypo:
> I had a small issue with the benchmark: a sample takes 8 seconds

The mdp benchmark is not run by default, only when "-b all" or "-b mdp" is used. performance 0.4.1 now includes also all benchmarks but pybench and mdp in the default group.

PR #17, opened 2016-10-18
Add regex_dna benchmark

haypo:
Patch written by Serhiy Storchaka: http://bugs.python.org/issue26436

It seems like Pyston has the same benchmark with minor changes:
* https://github.com/dropbox/pyston-perf/blob/master/benchmarking/benchmark_suite/fasta.py
* Pyston benchmark is called "fasta"
* fasta replaces stdout to check the output
* fasta moved the LCG RNG to a genRandom() subfunction which is a good idea to make random_fasta() more readable. Since it's only used to initialize the benchmark, it has no impact on performance.
* probably other subtle differences, I didn't check carefully

### DIFF
--- a/performance/benchmarks/bm_regex_dna.py
+++ b/performance/benchmarks/bm_regex_dna.py
@@ -1,0 +1,237 @@
+#!/usr/bin/env python
+"""
+The Computer Language Benchmarks Game
+http://benchmarksgame.alioth.debian.org/
+
+regex-dna Python 3 #5 program:
+contributed by Dominique Wahli
+2to3
+modified by Justin Peel
+
+fasta Python 3 #3 program:
+modified by Ian Osgood
+modified again by Heinrich Acker
+modified by Justin Peel
+Modified by Christopher Sean Forgeron
+"""
+
+import bisect
+import re
+
+import perf
+from six.moves import xrange
+
+
+DEFAULT_INIT_LEN = 100000
+DEFAULT_RNG_SEED = 42
+
+ALU = ('GGCCGGGCGCGGTGGCTCACGCCTGTAATCCCAGCACTTTGG'
+       'GAGGCCGAGGCGGGCGGATCACCTGAGGTCAGGAGTTCGAGA'
+       'CCAGCCTGGCCAACATGGTGAAACCCCGTCTCTACTAAAAAT'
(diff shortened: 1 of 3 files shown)


serhiy-storchaka:
"fasta" is used just for generating input data for regex-dna. The implementation in my patch is hardly optimized, since the initialization time is not such small.

performance/benchmarks/bm_regex_dna.py
```diff
@@ -0,0 +1,213 @@
...
+    is_trailing_line = False
+    count_modifier = 0.0
+
+    line = bytearray(width + 1)    # Width of 60 + 1 for the \n char
+
+    probs, chars = make_cumulative(table)
+
+    # pRNG Vars
+    im = 139968.0
+    #seed = 42.0
```
serhiy-storchaka: This can be removed.

haypo: The variable is still used. Do you suggest to replace the variable by its value in the code?

haypo:
> The implementation in my patch is hardly optimized, since the initialization time is not such small.

Oh right, it takes ~300 ms on CPython 3.5 on my laptop. By the way, I "optimized" the benchmark to only call init_benchmark() once. Before, it was initialized once per sample: perf uses 4 samples per process by default (including 1 warmup sample).

haypo:
I merged the pull request. We still can push minor enhancements later, like the comment on init_benchmarks(). Thanks Serhiy!

One less bechmark to port from Pyston (in Pyston, the benchmark is called "fasta").

serhiy-storchaka:
> One less bechmark to port from Pyston (in Pyston, the benchmark is called
> "fasta").

The "fasta" benchmark and the "regex-dna" benchmark are *different* benchmarks.

serhiy-storchaka:
> +    #seed = 42.0
> 
> The variable is still used. Do you suggest to replace the variable by its
> value in the code?

I meant only the commented out assignment.

haypo:
> The "fasta" benchmark and the "regex-dna" benchmark are *different* benchmarks.

Oops, I fixed the TODO document.

> I meant only the commented out assignment.

Ah. I replaced the constant with a command line option which has a default value of 42 :-)

haypo:
I consider that the issue (pull request) is done. If you have more remarks, you might open a new issue or send a new pull request.

PR #19, opened 2016-12-03
Add python_startup_stdlibs benchmark

methane:
There are some very common stdlib modules which is not imported by site module.
They affects startup time of command line applications.

Having benchmark to import them helps we notice how stdlibs going fatter.

### DIFF
--- a/performance/benchmarks/bm_python_startup.py
+++ b/performance/benchmarks/bm_python_startup.py
@@ -8,6 +8,33 @@ import perf
 from six.moves import xrange
 
 
+# common stdlibs. Many applications imports them directly, or indirectly.
+IMPORT_COMMON_STDLIBS = """\
+import abc
+import argparse
+import collections
+import copy
+import ctypes
+import datetime
+#import enum  # Python >=3.4
+import functools
+import io
+import json
+import logging
+import os
+#import pathlib  # Python >=3.4
+import re
+import socket
+import struct
+import subprocess
+import tempfile
+import threading
+import types
+#import typing  # Python >=3.5
+import urllib
+"""
+
+
 def bench_startup(loops, command):
     run = subprocess.check_call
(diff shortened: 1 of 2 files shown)


haypo:
I like the principle of the benchmark, but I dislike the "python_startup" time. Maybe you can add a new "import_time" benchmak?

About the modules only available in some Python versions, I dislike the idea of using the subset of Python 2, but I still want to compare Python 2 to Python 3.

I propose to add an argument: the minimum Python version. For example, your proposed benchmark would be import_time_py27. For me, there are two main usages of benchmarks: compare Python 3 to Python 2, and check for regressions in the development version of Python. So maybe we can only use two benchmarks:

* import_time_py27: imports for Python >= 2.7
* import_time_py36: imports for Pyhon >= 3.6

Since Python 3.6 will be the new stable benchmarks in a few weeks. For example, Python 3.6 has a new secrets module.

performance/benchmarks/bm_python_startup.py
```diff
@@ -8,6 +8,33 @@
 from six.moves import xrange
 
 
+# common stdlibs. Many applications imports them directly, or indirectly.
```
haypo: Can you please elaborate how you selected the "common stdlibs"?

ncoghlan: I'm not sure how @methane did it, but the public GitHub data set on Big Query can be a useful proxy for module popularity: https://cloud.google.com/bigquery/public-data/github

While it misses a lot of code (both proprietary and open source), it's still likely to be more representative than our own intuition.

haypo:
Since the purpose of this benchmark is unclear and that it seems to behave differently depending on the Python benchmark, I reject this PR.

I would prefer to only have benchmarks which behave the same on all Python implementation.

PR #20, opened 2017-01-02
Rename --with-optimization to --enable-optimizations

vincentfretin:
The option --with-optimization was renamed to --enable-optimizations, see https://bugs.python.org/issue26359#msg281301


berkerpeksag:
Thanks!

PR #30, opened 2017-08-09
adds threading benchmark

bhavishyagopesh:
It adds the threading bencmark for an initial case.I will add benchmark  for `Locks`  or other things(as possibly suggested by others).

### DIFF
--- a/performance/benchmarks/bm_threading.py
+++ b/performance/benchmarks/bm_threading.py
@@ -1,0 +1,33 @@
+"""
+Benchmark threading module.
+"""
+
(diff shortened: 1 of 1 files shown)


haypo:
I don't understand the purpose of the benchmark. Does it benchmark sleep? :-)

bhavishyagopesh:
That was a placeholder, now I have made the actual changes.

haypo:
Again, I don't understand well what do you want to measure. I'm not sure that such microbenchmark belongs to this benchmark suite. Maybe write to the speed mailing list to explain your rationale.

PR #32, opened 2017-10-13
Inherit the environment in recursive invocations

pdox:
On my firewalled box, the only way for pip to succeed is by having the the http_proxy / https_proxy environment variables set properly. Unfortunately, --inherit-environ option wasn't working as expected. The command I used was:

./pyperformance compile --inherit-environ=http_proxy,https_proxy benchmark.conf $mybranch

This would eventually invoke pip without the http_proxy/https_proxy environment variables set, causing it to timeout and fail. This commit fixes the problem.

Feel free to modify, and I will test it again on my side.

### DIFF
--- a/performance/compile.py
+++ b/performance/compile.py
@@ -126,8 +126,9 @@ class Repository(Task):
 
 
 class Application(object):
-    def __init__(self, conf):
+    def __init__(self, conf, options):
         self.conf = conf
+        self.options = options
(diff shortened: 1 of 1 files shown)


haypo:
Oh thanks for this nice bug fix!

Would you mind to document this change in doc/changelog.rst? You will have to add a new "Version 0.6.1" section. Create a new PR for that.

PR #36, opened 2018-03-20
COMPAT: PyPy does not have c-accelerators for pickle

mattip:
change the logic to detect/use c-accelerators for pickle when using PyPy for python 3.5 (called PyPy3 by the PyPy people)

### DIFF
--- a/performance/benchmarks/bm_pickle.py
+++ b/performance/benchmarks/bm_pickle.py
@@ -17,12 +17,14 @@ from __future__ import division
 import datetime
 import random
 import sys
+import platform
 
 import perf
 import six
 from six.moves import xrange
 if six.PY3:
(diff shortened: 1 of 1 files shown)


performance/benchmarks/bm_pickle.py
```diff
@@ -17,12 +17,14 @@
...
 import random
 import sys
+import platform
 
 import perf
 import six
 from six.moves import xrange
 if six.PY3:
     long = int
+IS_PYPY = platform.python_implementation() == 'PyPy'
```
vstinner: Can you please use perf.python_implementation() == 'pypy'?

vstinner:
Thank you! I merged your PR.

PR #43, opened 2018-07-25
Fix --inside-venv boolean flag to actually work

VictorRodriguez:
By enablement of this bool flag, the performance benchmark first creates a
virtual environment where it installs requirements with fixed versions to get
a reproducible environment. If this flag is not set the benchmark takes
the installed requirements by OS provider. Before this patch, the default was to create the virtual environment as default

Signed-off-by: Victor Rodriguez <victor.rodriguez.bahena@intel.com>

### DIFF
--- a/performance/cli.py
+++ b/performance/cli.py
@@ -202,7 +202,7 @@ def _main():
         cmd_show(options)
         sys.exit()
 
-    if not options.inside_venv:
+    if options.inside_venv:
         exec_in_virtualenv(options)
 
     from performance.cli_run import cmd_run, cmd_list, cmd_list_groups


vstinner:
The CI failed on your PR. I'm not sure that the change is correct. Can you try to have a look? Sorry, I missed this PR.

VictorRodriguez:
No problem @vstinner, I can take a look, however, I think the CI/CD fails due to the nature of the change. The reason I created this PR was to change the default set up of the benchmark. Right now it first creates a virtual environment where it installs requirements with fixed versions to get
a reproducible environment. What I am looking for is the cases when the community wants to benchmark the performance of python with the requirements provided by OS ( in order to benchmark python among multiple OS providers). Let me know if you like this proposed change.

vstinner:
In the meanwhile, I close the PR.

PR #50, opened 2019-04-11
Add support for native libraries that are installed but not on path

apparebit:
At least on macOS, installing up-to-date versions of several libraries
via a third party package manager such as Homebrew may interfere with
the rather old and tired version shipped with macOS. Consequently,
Homebrew does not add these libraries to the include and linker paths.
This change adds a configuration option to derive the correct arguments
for CPython's configure script. Adding support for other package
managers and operating systems should be trivial, since only two lines
are macOS-specific.

### DIFF
--- a/doc/benchmark.conf.sample
+++ b/doc/benchmark.conf.sample
@@ -44,7 +44,11 @@ pgo = True
 # to configure. As an exception, the prefix for openssl, if that
 # library is present here, is passed via the --with-openssl
 # option. Currently, this only works with Homebrew on macOS.
-pkg_only = openssl zlib
+# If running on macOS with Homebrew, you probably want to use:
+#     pkg_only = openssl readline sqlite3 xz zlib
+# The version of zlib shipping with macOS probably works as well,
+# as long as Apple's SDK headers are installed.
+pkg_only =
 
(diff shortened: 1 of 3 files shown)


doc/benchmark.conf.sample
```diff
@@ -37,6 +37,15 @@ lto = True
...
 pgo = True
 
+# The space-separated list of libraries that are package-only,
+# i.e., locally installed but not on header and library paths.
+# For each such library, determine the install path and add an
+# appropriate subpath to CFLAGS and LDFLAGS declarations passed
+# to configure. As an exception, the prefix for openssl, if that
+# library is present here, is passed via the --with-openssl
+# option. Currently, this only works with Homebrew on macOS.
+pkg_only = openssl zlib
```
DinoV: Should this be commented out in the default sample given that it will error out on non-Mac OS/X platforms?

apparebit: I considered that but decided to leave it in, since other options also have concrete values. But after going through the options again, I agree with you. Good catch!

Plus the actual list of libraries to make it all work on macOS is a tad longer, at least in the setup that worked for me. I'll run a few more experiments to hopefully find a minimal list and update the example, out-commenting it as well.

Issue #76, opened 2019-12-17
Django and Tornado benchmarks broken on Python 3.9


vstinner:
See also issue #74.

```
2019-12-17 01:04:41,423: [42/47] tornado_http...
2019-12-17 01:04:41,424: INFO:root:Running `/home/haypo/bench_tmpdir/venv/bin/python -u /home/haypo/pyperformance/pyperformance/benchmarks/bm_tornado_http.py --verbose --output /tmp/tmp8_0ujs9g`
2019-12-17 01:04:41,537: Traceback (most recent call last):
2019-12-17 01:04:41,537:   File "/home/haypo/pyperformance/pyperformance/benchmarks/bm_tornado_http.py", line 15, in <module>
2019-12-17 01:04:41,537:     from tornado.httpclient import AsyncHTTPClient
2019-12-17 01:04:41,537:   File "/home/haypo/bench_tmpdir/venv/lib/python3.9/site-packages/tornado/httpclient.py", line 50, in <module>
2019-12-17 01:04:41,537:     from tornado import gen, httputil, stack_context
2019-12-17 01:04:41,537:   File "/home/haypo/bench_tmpdir/venv/lib/python3.9/site-packages/tornado/httputil.py", line 107, in <module>
2019-12-17 01:04:41,537:     class HTTPHeaders(collections.MutableMapping):
2019-12-17 01:04:41,537: AttributeError: module 'collections' has no attribute 'MutableMapping'
```

and

```
2019-12-17 00:43:47,774: [ 6/47] django_template...
2019-12-17 00:43:47,775: INFO:root:Running `/home/haypo/bench_tmpdir/venv/bin/python -u /home/haypo/pyperformance/pyperformance/benchmarks/bm_django_template.py --verbose --output /tmp/tmp32v4efa1`
2019-12-17 00:43:47,925: Traceback (most recent call last):
2019-12-17 00:43:47,925:   File "/home/haypo/pyperformance/pyperformance/benchmarks/bm_django_template.py", line 38, in <module>
2019-12-17 00:43:47,925:     django.setup()
2019-12-17 00:43:47,925:   File "/home/haypo/bench_tmpdir/venv/lib/python3.9/site-packages/django/__init__.py", line 18, in setup
2019-12-17 00:43:47,925:     from django.urls import set_script_prefix
2019-12-17 00:43:47,925:   File "/home/haypo/bench_tmpdir/venv/lib/python3.9/site-packages/django/urls/__init__.py", line 1, in <module>
[...]

PR #77, opened 2019-12-17
Upgrade bm_tornado_http to Tornado 6.0.3

vstinner:


vstinner:
https://github.com/python/pyperformance/issues/76 becomes a blocker issue for Python 3.9, so I propose to upgrade to Tornado 6 and abandon Python 2.7 support.

I'm not sure about `yield self.flush()`, but I don't get any warning, so it might be ok :-)

vstinner:
I rebased the PR on top of #78 which dropped Python 2 support.

PR #81, opened 2020-03-30
Enable running with alternative python, inside venv

tomerv:
I wanted to run pyperformance benchmark with another executable, while using a venv for the pyperformance and benchmark packages. I had to make these changes to get it to work.

### DIFF
--- a/pyperformance/cli.py
+++ b/pyperformance/cli.py
@@ -167,7 +167,7 @@ def parse_args():
     if hasattr(options, 'python'):
         # Replace "~" with the user home directory
         options.python = os.path.expanduser(options.python)
-        # Try to the absolute path to the binary
+        # Try to get the absolute path to the binary
         abs_python = os.path.abspath(options.python)
         if not abs_python:
             print("ERROR: Unable to locate the Python executable: %r" %
--- a/pyperformance/cli.py
+++ b/pyperformance/cli.py
@@ -150,9 +150,7 @@ def parse_args():
                                "inside the virtual environment."))
         cmd.add_argument("-p", "--python",
                          help="Python executable (default: use running Python)",
-                         # See comment below
-                         # default=sys.executable
-                         )
+                         default=sys.executable)
         cmd.add_argument("--venv",
                          help="Path to the virtual environment")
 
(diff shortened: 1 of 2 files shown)


vstinner:
Hi. How do you reproduce your issue? What is your OS and Python version? Is it CPython?

tomerv:
It's a corporate Linux installation based on SuSE. I created a venv with Python 3.7.4, then tried to run pyperformance (from the venv) with this command:
``./venv/bin/python -m pyperformance run -f --venv ./venv --python tester -o output/tester.json``
Where ``tester`` is a script that calls my modified python version (actually it's more complicated than that, but I don't think that's important for now). Before I made those changes it didn't reach the point where ``tester`` is called.

pyperformance/cli.py
```diff
@@ -165,15 +167,23 @@ def parse_args():
...
+                print("ERROR: Unable to locate the Python executable: %r" %
+                      options.python)
+                sys.exit(1)
+            options.python = os.path.realpath(abs_python)
+        else:
+            # It would seems like it's possible to put sys.executable as the
+            # default when calling add_argument, but for some reason it returns
+            # another value in some cases: When the python binary is a symbolic
+            # link, the sys.executable in add_argument returns the realpath, while
+            # here it returns the relative path (before following the link).
```
vstinner: I don't understand why this change is needed. sys.executable should not be changed between line 151 and line 186.

tomerv: You're right, the comment I wrote is incorrect. I guess I got confused with the ``if hasattr(options, 'python')`` part.

The issue I have is with the line ``options.python = os.path.realpath(abs_python)`` - this follows the symbolic link from ``venv/bin/python`` to outside the venv.

My modification allows me to run without passing ``--python`` and then it goes into the ``else`` clause which doesn't have the ``realpath`` call.  I can't get around that without modifying the code in some way.

Is there a reason for this ``realpath`` call? Can it be removed? The other lines (replacing "~" and finding the absolute path) are not an issue. I will update the review accordingly.

pyperformance/cli_run.py
```diff
@@ -26,11 +26,14 @@ def cmd_run(parser, options):
         print("ERROR: the output file %s already exists!" % options.output)
         sys.exit(1)
 
-    if not os.path.isabs(sys.executable):
-        print("ERROR: sys.executable is not an absolute path")
+    executable = sys.executable
+    if hasattr(options, 'python'):
+        executable = options.python
```
vstinner: nitpick:

```suggestion
    if hasattr(options, 'python'):
        executable = options.python
    else:
        executable = sys.executable
```

pyperformance/cli.py
```diff
@@ -168,12 +168,12 @@ def parse_args():
         # Replace "~" with the user home directory
         options.python = os.path.expanduser(options.python)
         # Try to the absolute path to the binary
```
vstinner:
```suggestion
        # Try to get the absolute path to the binary
```

pyperformance/cli.py
```diff
@@ -167,13 +167,13 @@ def parse_args():
     if hasattr(options, 'python'):
         # Replace "~" with the user home directory
         options.python = os.path.expanduser(options.python)
-        # Try to the absolute path to the binary
-        abs_python = shutil.which(options.python)
+        # Try to get the absolute path to the binary
+        abs_python = os.path.abspath(options.python)
```
vstinner, line 171: Yeah, I recall that I also had issues when I resolved symlinks of env/bin:python. I don't recall when this binary is a symlink or not. I just recall that it should not always resolve symlinks ;-)

vstinner:
Thank you for the fix! I merged your PR.

PR #86, opened 2021-01-15
Add benchmark reading instance variables defined by __slots__

gvanrossum:
(Copied some parts from bm_float.py)


vstinner (commented):
I'm not really excited by micro-benchmarks like "read an atribute". Such micro-benchmark is usually unstable and so not very useful to track Python performance over time. We tried to put "macro" benchmarks on pyperformance, more "realistic" "real world" work loads (please don't look into details ;-)).

gvanrossum:
No problem, this was suggested to me. I'll withdraw it.

Issue #89, opened 2021-03-18
Add a way to plug in custom benchmarks.


ericsnowcurrently:
Currently you have to modify this repo if you want to run a custom benchmark.  It would be nice to have a mechanism by which a custom benchmark could be plugged in externally.

(This isn't a priority.)

vstinner:
pyperformance is just a set of benchmarks written with pyperf and runs them sequentially using --append option to add all results into an unique JSON file. Most of the work to manage benchmark results is done by pyperf. See pyperf --append option:
https://pyperf.readthedocs.io/en/latest/runner.html#json-output

You can run manually your custom benchmark with --append. It's common that I run benchmarks manually:

```
# Benchmark Python 3.9
$ python3.9 bm_telco.py --append py39.json
.....................
telco: Mean +- std dev: 8.16 ms +- 0.39 ms

$ python3.9 bm_unpack_sequence.py --append py39.json
.....................
unpack_sequence: Mean +- std dev: 80.6 ns +- 3.2 ns

$ python3.9 -m pyperf show py39.json 
telco: Mean +- std dev: 8.16 ms +- 0.39 ms
unpack_sequence: Mean +- std dev: 80.6 ns +- 3.2 ns


# Benchmark Python 3.6
$ python3.6 bm_telco.py --append py36.json
.....................
telco: Mean +- std dev: 10.2 ms +- 0.3 ms

$ python3.6 bm_unpack_sequence.py --append py36.json
.....................

$ python3 -m pyperf compare_to py36.json py39.json  
telco: Mean +- std dev: [py36] 10.2 ms +- 0.3 ms -> [py39] 8.16 ms +- 0.39 ms: 1.26x faster
unpack_sequence: Mean +- std dev: [py36] 72.2 ns +- 1.8 ns -> [py39] 80.6 ns +- 3.2 ns: 1.12x slower

Geometric mean: 1.06x faster

$ python3 -m pyperf compare_to py36.json py39.json  --table
+-----------------+---------+-----------------------+
| Benchmark       | py36    | py39                  |
+=================+=========+=======================+
| telco           | 10.2 ms | 8.16 ms: 1.26x faster |
+-----------------+---------+-----------------------+
| unpack_sequence | 72.2 ns | 80.6 ns: 1.12x slower |
+-----------------+---------+-----------------------+
| Geometric mean  | (ref)   | 1.06x faster          |
+-----------------+---------+-----------------------+
```

[...]

PR #90, opened 2021-03-18
Add a benchmark for the Azure CLI.

ericsnowcurrently:
This is part of CPython performance work that @gvanrossum and I are starting.  We want to use pyperformance to run our benchmarks (and profiling, but that's a separate issue).  As there isn't a mechanism to plug in custom benchmarks, I'm putting up this PR.

I suppose there's the question of whether this is an appropriate benchmark for the suite.  I think so.  It represents a workload, based on concrete usage patterns for a widely used CLI app, that does not seem to be covered by other benchmarks.  Note that at the moment I'm cheating a little by running tests out of the azure-cli test suite.  However, I expect the tests exercise the code in a very similar way to common usage of the actual application.  If there are concerns about inclusion I'm fine with leaving it out.

We will be using this benchmark (and a few others we will probably add) extensively.

(Either way, it would still be nice to have a way to plug in custom benchmarks externally.  See #89.)


ericsnowcurrently:
Hmm...

```
Execute: 'C:\Users\RUNNER~1\AppData\Local\Temp\tmpg12c4haj\Scripts\python.exe' -m azdev setup --cli 'D:\a\pyperformance\pyperformance\pyperformance\benchmarks\data\azure-cli'
Fatal Python error: _Py_HashRandomization_Init: failed to get random numbers to initialize Python
Python runtime state: preinitialized
```

ericsnowcurrently:
FYI, using azure-cli on 3.9+ is currently blocked: https://github.com/Azure/azure-cli/issues/16611.

PR #91, opened 2021-03-24
requirements.txt: bump up pyperf into 2.2.0

corona10:


vstinner:
Previously, I tried to keep track of every dependency upgrade in doc/changelog.rst, but it's painful to do it manually. So I just wrote a generic "Update requirements."

PR #95, opened 2021-04-24
Add an example for compile_all_revisions

### LINKED ISSUE #94: [Docs] Example for compile_all_revisions in config file
Hi all,

Recently I've been trying to benchmark a commit in cpython ``master`` branch against another commit in the same branch. The docs provide an example ``doc/benchmark.conf.sample`` file and describe it as 

> [compile_all_revisions]
> list of 'sha1=' (default branch: 'master') or 'sha1=branch'

I spent a few hours digging through the ``compile.py`` code to figure out what I was doing wrong before I finally realized I misinterpreted the docs. So I'd like to submit a PR to update the docs to make it slightly less confusing.

Thanks for your time!

Fidget-Spinner:
Fixes #94

### DIFF
--- a/doc/benchmark.conf.sample
+++ b/doc/benchmark.conf.sample
@@ -94,3 +94,5 @@ branches = default 3.6 3.5 2.7
 [compile_all_revisions]
 # list of 'sha1=' (default branch: 'master') or 'sha1=branch'
 # used by the "pyperformance compile_all" command
+# e.g.:
+11159d2c9d6616497ef4cc62953a5c3cc8454afb =


pablogsal (approved):

pablogsal:
Thanks @Fidget-Spinner !

vstinner:
I would prefer to put the example in a comment, so the configuration sample is directly usable.

Fidget-Spinner:
> I would prefer to put the example in a comment, so the configuration sample is directly usable.

The hash is a real one from cpython master branch, so it runs, but I get what you mean. Would you like me to make a PR to fix that?

vstinner:
> Would you like me to make a PR to fix that?

Hum. I looked at the rendered documentation: https://pyperformance.readthedocs.io/usage.html#compile-python-to-run-benchmarks

I forgot that the "sample" configuration already contains other values like branch names. So it's ok to keep this change without putting it into a comment.

Thanks @Fidget-Spinner for this nice enhancement ;-)

Issue #96, opened 2021-04-26
bm_pickle.is_module_accelerated() is named confusingly


gvanrossum:
The function returns True if the module is *not* accelerated. Either it should be renamed, or the sense of the return value should be inverted; either way both call sites should be updated, because currently they read backwards.

PR #97, opened 2021-04-26
bm_pickle: Rename is_module_accelerated to is_accelerated_module

corona10:
closes: https://github.com/python/pyperformance/issues/96

### DIFF
--- a/pyperformance/benchmarks/bm_pickle.py
+++ b/pyperformance/benchmarks/bm_pickle.py
@@ -236,8 +236,8 @@ BENCHMARKS = {
 }
 
 
-def is_module_accelerated(module):
-    return getattr(pickle.Pickler, '__module__', '<jython>') == 'pickle'
+def is_accelerated_module(module):
+    return getattr(pickle.Pickler, '__module__', '<jython>') != 'pickle'
 
 
 def add_cmdline_args(cmd, args):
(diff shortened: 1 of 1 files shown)


gvanrossum (approved):
Thanks for the quick turnaround! I hope you have write permission in this repo (AFAIK I don't).

vstinner:
English is hard, thanks for the fix @corona10 ;-)

gvanrossum:
I'm sorry, I have a follow-up. The help text for the `--pure-python` flag is backwards (it says "Use the C version of pickle" but it has the opposite effect). I'm beginning to think that at some point in the past the flag's sense was the opposite, and with Python 3 it was renamed, but the help text was not fixed. :-)

Finally, it looks like recent commits are duplicated, with one commit that does the work followed by a commit that touches no files by @pablogsal. What's up with that? I didn't even know you could do an empty commit. :-)

pablogsal:
> to always squash-merge

Good point. That has to be configured by @vstinner as I don't have admin to this repo.

gvanrossum:
Maybe the repo needs more than one owner? Redundancy! (Not that I'm
volunteering. :-)

vstinner:
I changed @methane permission to Admin and I invited @pablogsal to give him admin permission as well.

I don't know how to change that.

pablogsal:
> I don't know how to change that.

I changed it

PR #109, opened 2021-10-04
Support user-defined benchmark suites.

ericsnowcurrently:
(I realize this is a big PR.  Sorry!  The changes to .py files is a smaller change.  If necessary I can split it up a little, but it might not be worth it. 🙂  Just let me know.)

Currently pyperformance is two things coupled together: a tool to run a Python benchmark suite and a curated suite of Python benchmarks.  This PR splits those apart, with the existing suite used as the default.  This allows users to run their own set of benchmarks, perhaps specific to their Python implementation or their PyPI library, e.g. https://github.com/ericsnowcurrently/pyston-macrobenchmarks/tree/pyperformance.

Key changes:

* introduce a new filesystem structure for suites and individual benchmarks
* add the `--manifest` CLI option to specify the custom suite to use
* the default suite has been changed to the new format and moved to pyperformance/_benchmarks (only a data dir now)
* sometimes run some benchmarks in separate venvs
* do not fail all benchmarks if the dependencies for one cannot be installed

Most notably, this change should not affect benchmark results.

## Specifying a Benchmark Suite

A benchmark suite is defined through a manifest file.  For example, see the file for the default suite in `pyperformance/_benchmarks/MANIFEST`.  A manifest file has a `[benchmarks]` section (and zero or more `[group NAME]` sections).  See pyperformance/benchmarks/_manifest.py.

The `[benchmarks]` section is a TSV (tab-separated-values) table with 4 columns:

* name - the name of the benchmark
* version - (optional) the version of that benchmark to use
* origin - (optional) the URI where the benchmark is found (e.g. PyPI, GitHub, local filesystem)
* metafile - the location of the metadata file for the benchmark (or `<local>` to find "bm_<name>" relative to the manifest file)

## The New Structure of a Benchmark

Each benchmark is in its own directory and looks like this:

```
<benchmark>/
    pyproject.toml
    bm_<name>.toml    # alternate benchmarks based on this one
    requirements.txt  # if has external dependencies; pins versions
    run_benchmark.py  # essentially the same benchmark script as before
```

`pyproject.toml` is the normal [PEP 621](https://www.python.org/dev/peps/pep-0621/) format for Python projects.  These are the important fields in the `[project]` section:

* "name" - (may be inferred from the suite manifest)
* "version" - (may be inferred from the suite manifest or base metadata)
* "dependencies" - (if has external dependencies; the top-level list of dependencies)

We add a `[tool.pyperformance]` section, as supported by [PEP 518](https://www.python.org/dev/peps/pep-0518/#tool-table).  These are the important fields:

* "name" - (inferred from the `[project]` section)
* "metabase" - if any, the pyproject.toml to inherit from (".." expands to "../base.toml"; possibly inferred)
* "extra_opts" - if any, additional CLI args to use when running the benchmark script
* "runscript" - defaults to "./run_benchmark.py"

See pyperformance/benchmark/_metadata.py.


ericsnowcurrently:
Thanks for the helpful comments, @vstinner!

> * Why is there now **3** benchmark directories? `ls -d *benchmark*` gives `benchmark/  _benchmarks/  benchmarks/`. The purpose of each directory is not obvious to me.

pyperformance/benchmark/*.py - code about a single benchmark
pyperformance/benchmarks/*.py - code about a set of benchmarks
pyperformance/_benchmarks/ - the default benchmarks

> Why not keeping benchmarks... in the benchmarks/ directory?

I'll fix this so it's more obvious where the benchmarks are.

> * I suggest to keep all Python files in `pyperformance/*.py` and avoid nested sub-directories.

I'm fine with that.  It just means there will be a lot of files at the top-level.
 
> Most TOML files contains these lines:
> ...

Yeah, I haven't documented the format very well.  I'll explain it in a README.

> I'm not sure of the value of all these comments. Maybe just remove all of them?

I'll remove the unnecessary ones and explain any that are left.

> I don't understand `dynamic`, what is that? And what is `metabase = ".."`?

This will be explained in the README.

Basically, "dynamic" is part of the PEP 621 spec.  See https://www.python.org/dev/peps/pep-0621/#dynamic.

"metabase" is the benchmark .toml file to inherit from.  That way you don't have to duplicate metadata.  That name isn't great though.  I'll fix it.  Also, duplication may not be a big deal since the metadata isn't that complex.  It might make sense getting rid of the idea of inheriting metadata, to simplify things.  However, if you have lots of benchmarks then inheritance is really helpful when you have to update any common metadata.  (With inheritance you update the one file.)

> pyperformance/_benchmarks/bm_pickle/ contains 7 TOML files (total: 91 lines) for a single benchmark (run_benchmark.py: 286 lines).

There's one .toml file for each of the pickle benchmarks, with one of them the "base" spec.  They share the one Python script.

> It sounds quite complicated.

This will be explained in the README.  That said, if this approach (to multiple benchmarks sharing a script) isn't obvious, and clearly it wasn't, then we should find a better way.  Perhaps make the filenames indicate that they are variants?  Do that in the metadata?  Maybe allow multiple `[pyperformance]` entries in the .toml file instead of having a separate file for each variant?

I'll look into this.

> Maybe the Python script should be modified to run the 7 benchmarks at once, and maybe have opt-in option to only run a single benchmark, rather than the opposite?

Yeah, that's a possibility.  However, there is definitely a use case for having multiple variants of a benchmark running separately.  It would be good to support that explicitly.

> Can't you put the version and origin in the TOML file of each benchmark, rather than having to put it in a MANIFEST file?

This will get explained in the README.

"origin" is there so a manifest can specify that a benchmark should get pulled from somewhere else (e.g. PyPI or GitHub).  This contrasts with "metafile", which indicates the location of the benchmark's local metadata file.  For the default benchmarks the benchmarks are already local so they don't have an "origin" entry.

"version" is there specify the exact version to get when pulling a benchmark from "origin".  This is important for pinning benchmarks to get consistent results.  It isn't set for the default benchmarks because their version is the version of pyperformance.

All that said, having "version" and "origin" in the manifest is only worth it if we provide a mechanism for downloading benchmarks.  No such mechanism exists, so we don't need them.  On top of that, even if there was such a mechanism, the manifest isn't the right place to specify that stuff.  We would probably add a separate pyperformance CLI command for downloading benchmarks and adding them to a manifest.

So I'm going to remove "origin" and "version".
 
> For benchmark groups, can't you store the groups in the TOML file?

Agreed.  I put the groups in the manifest file because it seemed sensible at the time.  However, I was considering just having a separate manifest file for each group instead.  Identifying groups in benchmark .toml files is a good idea too.

> IMO we should try to organize each benchmark as if it doesn't matter where it is on the filesystem.

Could you explain more what you mean?  That's pretty much the way it is now.  Each benchmark is self-contained.  The only exception is with "metabase".  Perhaps I misunderstood what you said?

ericsnowcurrently:
@vstinner, here's what I plan on doing in response to your comments:

* [ ] make it obvious where the default benchmarks are
* [ ] keep all source files in top-level directory
* [ ] drop unnecessary comments from the benchmark .toml files
* [ ] add a README to pyperformance/_benchmarks to explain the format
    * [ ] explain "dynamic" (in benchmark .toml files) in the new README
    * [ ] explain "metabase" in the README (or drop it)
    * [ ] explain the manifest "version" column (if kept)
    * [ ] explain the manifest "origin" column (if kept)
* [ ] rename "metabase"
* [ ] improve the story for benchmark variants (make inheritance more obvious?)
* [ ] combine the 7 pickle benchmarks?
* [ ] drop version / origin from manifests
* [ ] store groups in benchmark .toml files?

ericsnowcurrently:
FYI, I also need to rebase onto main.

ericsnowcurrently:
@vstinner, ping

ericsnowcurrently:
@pablogsal, if possible could you give this a quick look.  I don't necessarily need a full review of the code.  More than anything I want to be sure the overall approach is acceptable.  I'd be glad to hop into a call if that would help.  Thanks!

ericsnowcurrently:
FYI, after a quick chat with @pablogsal, I plan on merging this tomorrow.

PR #123, opened 2022-01-11
Remove unused var from fannkuch

carljm:
Not sure what the philosophy on changes to the benchmarks themselves is in this repo; the theory of this PR is that the benchmarks are intended to be reasonable representations of the most efficient way to accomplish a given (admittedly meaningless) task in Python with the correct result, and it doesn't make sense to have pure wasted work in the benchmarks.

This `check` var is initialized and incremented, but never used, and plays no role in returning the correct result.


ericsnowcurrently (approved):
+1

ericsnowcurrently:
Thanks, @carljm.

PR #124, opened 2022-01-12
Bump tested versions to Python 3.7 and 3.10

hugovk:
Python 3.10 has been released and 3.6 is now EOL, so bump the CI tests to run on 3.7 and 3.10 instead of 3.6 and 3.9.

Also:

* Add `workflow_dispatch` to GitHub Actions to add a button to allow manually triggering builds on demand
* No need to explicitly include 3.7/Ubuntu and 3.10/Ubuntu in the CI config, it's part of the matrix
* Remove the unused Azure Pipelines config, only GHA is used for CI, which tests Ubuntu and Windows
* Update examples in README


ericsnowcurrently:
Looks like this was done, at least in part, in #137.

hugovk:
Good stuff!

I've resolved the conflicts for the other bits and pieces, but feel free to close if you don't need it.

ericsnowcurrently:
Thanks.

PR #127, opened 2022-01-21
Include `requirements.txt` of benchmarks.

### LINKED ISSUE #126: `MANIFEST.in` lacks `bm_*/requirements.txt`.
**How to reproduce the issue**

Install a local pyperformance from another directory, e.g. `pip install ../pyperformance`,
then run benchmarks have additional dependencies, e.g. `pyperformance run -b chameleon`.
Run will fail and you can find there's no requirements.txt under installed `bm_chameleon/`.

This occurs to me when I'm using a forked of pyperformance to test my own package.
If folks are developing pyperformance itself, this won't be an issue because the repo root will be current directory and benchmarks in cwd, rather than the installed ones, will be used, so requirements can be found as expected.

PS: I'm using `pip install -e` as a workaround.

oraluben:
Closes #126.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -16,5 +16,6 @@ include pyperformance/data-files/requirements.txt
 include pyperformance/data-files/benchmarks/MANIFEST
 include pyperformance/data-files/benchmarks/bm_*/*.toml
 include pyperformance/data-files/benchmarks/bm_*/*.py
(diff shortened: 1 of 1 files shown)


ericsnowcurrently (approved):
LGTM

Thanks for the fix!

PR #128, opened 2022-01-22
refactoring code with Assign Multiple Targets

idiomaticrefactoring:
refactoring code with Assign Multiple Targets which is more pythonic, concise, readable and efficient; how do think this change which has practical value？

### DIFF
--- a/pyperformance/data-files/benchmarks/bm_scimark/run_benchmark.py
+++ b/pyperformance/data-files/benchmarks/bm_scimark/run_benchmark.py
@@ -343,12 +343,7 @@ def FFT_bitreverse(N, data):
         jj = j << 1
         k = n >> 1
         if i < j:
-            tmp_real = data[ii]
-            tmp_imag = data[ii + 1]
-            data[ii] = data[jj]
-            data[ii + 1] = data[jj + 1]
-            data[jj] = tmp_real
-            data[jj + 1] = tmp_imag
+            data[ii] , data[ii + 1] , data[jj] , data[jj + 1]  = data[jj], data[jj + 1], data[ii], data[ii + 1]
         while k <= j:
             j -= k
             k >>= 1


pyperformance/data-files/benchmarks/bm_scimark/run_benchmark.py
```diff
@@ -343,12 +343,7 @@ def FFT_bitreverse(N, data):
         jj = j << 1
         k = n >> 1
         if i < j:
-            tmp_real = data[ii]
-            tmp_imag = data[ii + 1]
-            data[ii] = data[jj]
-            data[ii + 1] = data[jj + 1]
-            data[jj] = tmp_real
-            data[jj + 1] = tmp_imag
+            data[ii] , data[ii + 1] , data[jj] , data[jj + 1]  = data[jj], data[jj + 1], data[ii], data[ii + 1]
```
ericsnowcurrently, line 346: Personally, I find that harder to read.

zjzh: How about adding a comment? it maybe more readable. 
# swap two imaginary numbers with real part and imaginary part  (data[ii],data[ii+1]) <->(data[jj],data[jj+1]) . 

Personally, it maybe more longer making it seems harder read. If we first look at data[ii] to the left of “=”, then look at data[jj] to the right of “=”, then loop the operation. It maybe more simpler to read. Sometimes, the original code have more statements making ourselves a little scared to read this code.

And it could improve little performance and memory. 
```
data=[i for i in range(10)]
ii=1
jj=3
code1='''
tmp_real = data[ii]
tmp_imag = data[ii + 1]
data[ii] = data[jj]
data[ii + 1] = data[jj + 1]
data[jj] = tmp_real
data[jj + 1] = tmp_imag
'''

code2='''
data[ii] , data[ii + 1] , data[jj] , data[jj + 1] = data[jj], data[jj + 1], data[ii], data[ii + 1]
'''
print(timeit.timeit(code1,number=1000,globals=globals()))#0.00032373203430324793
print(timeit.timeit(code2,number=1000,globals=globals()))#0.0004283020389266312
```

zjzh: How about adding a comment? it maybe more readable. 
`#swap two imaginary numbers with real part and imaginary part  (data[ii],data[ii+1]) <->(data[jj],data[jj+1]) . `

Personally, it maybe more longer making it seems harder read. If we first look at data[ii] to the left of “=”, then look at data[jj] to the right of “=”, then loop the operation. It maybe more simpler to read. Sometimes, the original code have more statements making ourselves a little scared to read this code.

And it could improve little performance and memory. 
```
data=[i for i in range(10)]
ii=1
jj=3
code1='''
tmp_real = data[ii]
tmp_imag = data[ii + 1]
data[ii] = data[jj]
data[ii + 1] = data[jj + 1]
data[jj] = tmp_real
data[jj + 1] = tmp_imag
'''

code2='''
data[ii] , data[ii + 1] , data[jj] , data[jj + 1] = data[jj], data[jj + 1], data[ii], data[ii + 1]
'''
print(timeit.timeit(code1,number=1000,globals=globals()))#0.00032373203430324793
print(timeit.timeit(code2,number=1000,globals=globals()))#0.0004283020389266312
```

kumaraditya303: It is indeed harder to read and we should avoid changing benchmarks unless necessary, BTW the bytecode will now use unpack sequence rather than store name so will have an effect on its specialization.

PR #132, opened 2022-02-08
Add simple memory allocation benchmark

tiran:


pyperformance/data-files/benchmarks/bm_alloc/run_benchmark.py
```diff
@@ -0,0 +1,56 @@
...
+
+This benchmark stresses memory allocator
+"""
+
+import pyperf
+
+
+ONE_MB = 1024 * 1024
+
+DEFAULT_SIZES = [32, 128, 512, 8192, ONE_MB, 8 * ONE_MB]
```
pablogsal, line 13: This looks good to me, but take into account that `obmalloc` will trigger only for blocks less than 512:

https://github.com/python/cpython/blob/b5527688aae11d0b5af58176267a9943576e71e5/Objects/obmalloc.c#L889

We could go as it is and test globally all allocators (but we would be benchmarking `malloc`, basically) or restrict this test to small allocations.

pyperformance/data-files/benchmarks/bm_alloc/run_benchmark.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def allocate(repeat, sizes, alloc_func=bytearray):
+    for size in sizes:
+        append = alloc_func(size)
+        for i in range(repeat):
+            # malloc()
+            block = alloc_func(size)
+            # realloc()
+            block += append
```
pablogsal, line 24: Nice!

pyperformance/data-files/benchmarks/bm_alloc/run_benchmark.py
tiran: I have updated the test to use smaller sizes in order to stress obmalloc. We can add another benchmark later to test libc malloc vs. mimalloc.

pablogsal:
I'm happy landing this. Let's give some days if @vstinner wants to take a look.

pyperformance/data-files/benchmarks/bm_alloc/run_benchmark.py
```diff
@@ -0,0 +1,73 @@
+# coding: utf-8
+"""
+Allocate blocks of memory with bytearray
+
+This benchmark stresses memory allocator. Default sizes are <=
+obmalloc small request threshold.
```
pablogsal, line 6:
```suggestion
This benchmark stresses the memory allocator. Default sizes are <=
obmalloc's small request threshold.
```

vstinner:
Does it belong to pyperformance which is supposed to be a set of "macro benchmarks"?

IMO it's to have benchmarks outside pyperformance. @ericsnowcurrently wants to add the ability to plug your own benchmark suite in pyperformance using configuration files, but I didn't follow his PR recently.

pablogsal:
> Does it belong to pyperformance which is supposed to be a set of "macro benchmarks"?

I would say yes: we have things that are quite small in what they do anyway. C'mon, we even have unpacking tests:

https://github.com/python/pyperformance/blob/034f58be960ee88010b7a8046265ca88f0c35aa4/pyperformance/data-files/benchmarks/bm_unpack_sequence/run_benchmark.py#L10-L20

vstinner:
> C'mon, we even have unpacking tests:

Maybe these tests don't belong to pyperformance :-(

> I would say yes: we have things that are quite small in what they do anyway. 

It's up to you.

ericsnowcurrently:
> Does it belong to pyperformance which is supposed to be a set of "macro benchmarks"?
> 
> @ericsnowcurrently wants to add the ability to plug your own benchmark suite in pyperformance

My PR landed a while back and is part of the most recent release.

That said, we already have a bunch of micro benchmarks so I'm not opposed to adding more.  (We can separate the bunch later.)  However, it would be worth marking them as "micro" sooner rather than later so they are easy to skip.  That would involve adding/updating `tags = "micro"` in the `[tool.pyperformance]` section of each benchmark's pyproject.toml.  For this PR it would probably be okay to do so just for the new benchmark.

pyperformance/data-files/benchmarks/bm_alloc/pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[project]
+name = "pyperformance_bm_alloc"
+requires-python = ">=3.8"
+dependencies = ["pyperf"]
+urls = {repository = "https://github.com/python/pyperformance"}
+dynamic = ["version"]
+
+[tool.pyperformance]
+name = "alloc"
```
ericsnowcurrently, line 9: Please tag it as a micro benchmark.

```suggestion
[tool.pyperformance]
name = "alloc"
tags = "micro"
```

ericsnowcurrently (changes requested):

pyperformance/data-files/benchmarks/bm_alloc/pyproject.toml
tiran: I have added the micro benchmark tag

pablogsal:
@tiran THe benchmark fails for pypy because `sys.getsizeof` doesn't work. Check the CI output

doc/benchmarks.rst
```diff
@@ -56,6 +53,18 @@ depending on the Python version.
    Files are called ``.py.txt`` instead of ``.py`` to not run PEP 8 checks on
    them, and more generally to not modify them.
 
+alloc
+-----
+
+Allocate blocks of memory with ``bytesarray``. The default sizes are smaller or
```
StanFromIreland:
```suggestion
Allocate blocks of memory with ``bytearray``. The default sizes are smaller or
```

PR #134, opened 2022-02-10
Fix EncodingWarning

methane:


ericsnowcurrently:
Thanks for fixing that!  Sorry I forgot to do it in the first place.

raghavthind2005 (approved):

PR #136, opened 2022-02-11
Replace `a = a + b` to `a += b` for strings

arhadthedev:
According to https://github.com/faster-cpython/ideas/discussions/269, concatenation of strings with writing a result back halts specialization of `BINARY_OP_INPLACE_ADD_UNICODE` opcode thus reducing performance.

In addition, this PR does the same replacement for numeric `lambertAmount` for the sake of consistency.


brandtbucher:
Thanks, but see my comment [here](https://github.com/faster-cpython/ideas/discussions/269#discussioncomment-2160686).

PR #140, opened 2022-02-24
Fix --inherit-environ issue

cappadokes:


pyperformance/data-files/requirements.txt
```diff
@@ -10,7 +10,7 @@ psutil==5.8.0
     # via -r requirements.in
 pyparsing==3.0.6
     # via packaging
-pyperf==2.3.0
+pyperf @ file:///home/cappadokes/code/pyperf
```
gvanrossum, line 13: This seems like something you changed locally, but that should be changed back before landing this.

gvanrossum (commented):
This PR's title doesn't match what the code does (it seems to be adding a `--track-energy` flag).

pyperformance/data-files/requirements.txt
cappadokes: Unfortunately this is somewhat tricky:
- for `--track-energy` to work, `pyperf` should also support energy measurements. There is no such `pyperf` release at the moment, but [we are working on it](https://github.com/psf/pyperf/pull/125).
- until then, if a user wants to use `--track-energy` from `pyperformance`, she must indeed use a local `pyperf` in `requirements.txt`.

cappadokes:
> This PR's title doesn't match what the code does (it seems to be adding a `--track-energy` flag).

To be precise, the code both fixes the `--inherit-environ` issue and it also adds a `--track-energy` flag. I apologize for not splitting this in 2 separate PRs, but the changes under discussion are partly interdependent (`--track-energy` needs `--inherit-environ` to work, though the converse does not hold).

I will soon follow up with a documentation commit for clarification. I will also change the PR's name so as to contain both points.

ericsnowcurrently:
@cappadokes, please put the `--track-energy` addition in a separate PR.  We will merge the fix for `--inherit-environ` first.

cappadokes:
> @cappadokes, please put the `--track-energy` addition in a separate PR. We will merge the fix for `--inherit-environ` first.

@ericsnowcurrently done. `--inherit-environ` is [here](https://github.com/python/pyperformance/pull/188) now.

PR #155, opened 2022-03-22
Share a venv, if possible

ericsnowcurrently:
This also includes a few small fixes.


oraluben:
Are there any risks that dependencies of benchmark a will affect the result of benchmark b when they share a venv? Or this will be resolved manually. e.g. `.pth` file or something like that.

ericsnowcurrently:
AFAIK, the only risk is if one benchmark modifies any of the dependencies, which shouldn't happen.

PR #163, opened 2022-03-26
Fix several failures.

ericsnowcurrently:
A few bugs have crept in with some of my recent changes.  Part of the problem is that there hasn't been great test coverage for a bunch of pyperformance code.  In this PR I've added a bunch of testing.  I've also fixed a number of problems.


gvanrossum:
I wonder how much of my Windows woes this fixes? I'll give it a try. @ericsnowcurrently please note that I *accidentally* pushed a change to greenlet==2.0.0a2. I meant to create a PR but due to my git clumsiness pushed to main. Sorry. But I think it's the right fix, so I hope we can keep it.

brandtbucher:
Just a note: you can use the `--no-cache-dir` option to ignore cached wheels, and `--force-reinstall` to skip the uninstall step. So...

```
.\python.bat -m pip install --no-cache-dir --force-reinstall greenlet==2.0.0a2
```

...should do it in one command.

gvanrossum:
Remembering those options might be harder than the original steps. :-)

Let's continue the discussion in https://github.com/faster-cpython/ideas/issues/333

PR #166, opened 2022-03-28
Preserve the Exit Code of Running the Tests

ericsnowcurrently:
We have been ignoring the exit code of the subprocess in runtests.py.

### DIFF
--- a/pyperformance/tests/test_commands.py
+++ b/pyperformance/tests/test_commands.py
@@ -1,7 +1,6 @@
 import os
 import os.path
 import shutil
-import subprocess
 import sys
 import textwrap
 import unittest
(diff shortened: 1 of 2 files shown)


ericsnowcurrently:
We will deal with failures separately.

brandtbucher:
It appears that this commit broke (or exposed a failure in?) our [PyPy CI job](https://github.com/python/pyperformance/runs/5727336762?check_suite_focus=true). Looks like venv creation isn't working right?

ericsnowcurrently:
Yeah, it is a pre-existing failure.  See https://github.com/python/pyperformance/issues/167.

PR #175, opened 2022-03-29
Add some notes about running on Windows

gvanrossum:
I figured I'd save future users some time when trying to use a python.exe built from source on Windows.


ericsnowcurrently (approved):
LGTM

Issue #139, opened 2022-02-24
run --inherit-environ option does not work


cappadokes:
After a clean install of `pyperformance`, run the below command:

```bash
HAHA=1 pyperformance run -f --debug-single-value --inherit-environ HAHA -b 2to3
```
The output received is:
```bash
[...]
[1/1] 2to3...
INFO:root:Running `$PATH_TO_VENV$/bin/python -u /usr/local/lib/python3.8/dist-packages/pyperformance-1.0.4-py3.8.egg/pyperformance/data-files/benchmarks/bm_2to3/run_benchmark.py --debug-single-value --inherit-environ =,PYPERFORMANCE_RUNID --output /tmp/tmpiq530yv4`
```
As you can see, the HAHA variable is not inherited by the benchmark-running process. Instead, a '=' symbol is parsed in its place.

cappadokes:
Update:

I have found the problem to be line 256 in pyperformance/_benchmark.py. The line is part of the `_resolve_restricted_opts` function:

```python
resolved.append(opt.partition('=')[-2]) 
```
Replacing -2 with -1 resolves the issue. I have created a relevant pull request (#140).

agraul:
I had the same problem and came up with the same fix. I can confirm it that this works.

PR #188, opened 2022-04-27
Fix --inherit-environ issue

cappadokes:
Problem initially appeared [here](https://github.com/python/pyperformance/issues/139). Solution introduced [here](https://github.com/python/pyperformance/pull/140), but was later requested to be split from that PR's main focus.


ericsnowcurrently (approved):
Thanks for doing this.

PR #187, opened 2022-04-27
Add benchmark for async tree workloads

arielin3:
### What

Add a benchmark for testing async workloads, specifically an async tree workload that simulates simpler versions of a typical Instagram endpoint. (_"typical implementation of endpoint in IG is an async function that in turn calls into numerous other async functions to generate an output."_ - see context below)

### Motivation/Context

See https://github.com/python/cpython/issues/91121

### Testing

Running ``pyperformance`` locally:

```
% python3 -m pyperformance list
'<default>' benchmarks:
- 2to3
- async_tree
- async_tree_cpu_io_mixed
- async_tree_io
- async_tree_memoization
- chameleon
- chaos
...
```

```
% python3 -m pyperformance run -o ~/python_efficiency_hackamonth/pyperformance_async_tree_output.json --benchmarks=async_tree,async_tree_cpu_io_mixed,async_tree_memoization,async_tree_io
Python benchmark suite 1.0.5

Creating the virtual environment /Users/arielin/python_efficiency_hackamonth/pyperformance_git/pyperformance/venv/cpython3.10-38b51ccaf402-compat-c0d88e07feb9

...

Performance version: 1.0.5
Python version: 3.10.4 (64-bit) revision 9d38120e33
Report on macOS-12.3.1-x86_64-i386-64bit
Number of logical CPUs: 16
Start date: 2022-05-04 18:06:49.877751
End date: 2022-05-04 18:13:51.586338

### async_tree_none ###
Mean +- std dev: 830 ms +- 59 ms

### async_tree_cpu_io_mixed ###
Mean +- std dev: 1.10 sec +- 0.05 sec

### async_tree_io ###
Mean +- std dev: 2.00 sec +- 0.05 sec

### async_tree_memoization ###
Mean +- std dev: 963 ms +- 37 ms
```

(EDIT: Testing updated on 5/4 after rename changes)

PR #190, opened 2022-05-02
Upgrade versions of deps

itamaro:
the motivation behind this is to bump pyperf to 2.3.1, which includes the new `Runner.bench_async_func()` API.
This API is required for the async benchmarks that @arielin3 added in PR #187


ericsnowcurrently (approved):

ericsnowcurrently:
Thanks, @itamaro!

PR #193, opened 2022-05-04
Treat non +- lines as positive

### LINKED ISSUE #191: Manifest "group" sections don't handle non-"+"/"-" lines well
I was writing a "group" section and I just wrote out the names of the benchmarks, but that crashed pyperformance with
```
  File "/home/kmod/pyston3/build/system_env/lib/python3.8/site-packages/pyperformance/_manifest.py", line 365, in _parse_group_section
    yield op, name
UnboundLocalError: local variable 'op' referenced before assignment
```

It looks like the non-`+`/`-` case doesn't set the variable `op`: https://github.com/python/pyperformance/blob/94310a38df07151e5059fc2974150f5e5adbb528/pyperformance/_manifest.py#L363

My assumption was that these should behave like `+` lines but I wasn't sure

kmod:
in manifest group sections

fixes #191

### DIFF
--- a/pyperformance/_manifest.py
+++ b/pyperformance/_manifest.py
@@ -360,6 +360,7 @@ def _parse_group_section(lines):
             op = '+'
             name = line[1:]
         else:
+            op = '+'
             name = line
         _benchmark.check_name(name)
         yield op, name


ericsnowcurrently (approved):

ericsnowcurrently:
Thanks @kmod!

PR #196, opened 2022-05-05
github: Update PyPy version from 3.6 to [3.7, 3.8, 3.9]

corona10:
default-pypy

Current GitHub Action only executes the default PyPy version of Ubuntu which is actually 3.6 and too old but we are executing the CI for CPython 3.[7-10].
We may need to expand the PyPy version to support their latest implementation and match the identical versions.


ericsnowcurrently (approved):

ericsnowcurrently:
Thanks, @corona10!

PR #197, opened 2022-05-10
Add benchmarks for ctypes function call overhead

mdboom:
This additionally adds support for building C extensions as part of creating a benchmark's virtual environment, since that didn't seem to be possible prior to this change.

See https://github.com/faster-cpython/ideas/issues/370 for some background discussion.


mdboom:
Wanted to add a note: This is definitely a "microbenchmark" that isn't a great example of a real-world workload.

pyperformance/data-files/benchmarks/bm_ctypes/run_benchmark.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def benchmark(n):
+    void_foo_void = ext.void_foo_void
+    int_foo_int = ext.int_foo_int
+    void_foo_int = ext.void_foo_int
+    void_foo_int_int = ext.void_foo_int_int
+    void_foo_int_int_int = ext.void_foo_int_int_int
+    void_foo_int_int_int_int = ext.void_foo_int_int_int_int
+    void_foo_constchar = ext.void_foo_constchar
```
ericsnowcurrently, line 24: For *micro* benchmarks using `runner.bench_func()` we want to keep the function body as tight as possible to the code we want to measure.   Depending on how fast each loop iteration is and what `n` is, it may make sense to move some of this out of the function:

```suggestion
_void_foo_void = ext.void_foo_void
_int_foo_int = ext.int_foo_int
_void_foo_int = ext.void_foo_int
_void_foo_int_int = ext.void_foo_int_int
_void_foo_int_int_int = ext.void_foo_int_int_int
_void_foo_int_int_int_int = ext.void_foo_int_int_int_int
_void_foo_constchar = ext.void_foo_constchar


def benchmark(n):
    void_foo_void = _void_foo_void
    int_foo_int = _int_foo_int
    void_foo_int = _void_foo_int
    void_foo_int_int = _void_foo_int_int
    void_foo_int_int_int = _void_foo_int_int_int
    void_foo_int_int_int_int = _void_foo_int_int_int_int
    void_foo_constchar = _void_foo_constchar
```
or:
```suggestion

def benchmark(n, *,
                         void_foo_void =ext.void_foo_void,
                         int_foo_int=xt.int_foo_int,
                         void_foo_int=ext.void_foo_int,
                         void_foo_int_int=ext.void_foo_int_int,
                         void_foo_int_int_int=ext.void_foo_int_int_int,
                         void_foo_int_int_int_int=ext.void_foo_int_int_int_int,
                         void_foo_constchar=ext.void_foo_constchar,
                         ):
```

But it's only worth it if ctypes is really fast.

Then again, the point is moot if you use `runner.bench_time_func()` instead of `runner.bench_func()`.

pyperformance/data-files/benchmarks/bm_ctypes/run_benchmark.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # Test calling the functions using the implied arguments mechanism
+
+    for i in range(n):
```
ericsnowcurrently, line 28: If we stick with `runner.bench_func()` and ctypes is fast enough, we have to factor in the overhead of the for loop.

pyperformance/_benchmark.py
```diff
@@ -164,6 +164,10 @@ def runscript(self):
     def extra_opts(self):
         return self._get_metadata_value('extra_opts', ())
 
+    @property
+    def setup_py(self):
+        return self._get_metadata_value('setup_py', False)
```
ericsnowcurrently, line 169: Let's return the filename (and check if the file exists) here:

```suggestion
    @property
    def setup_py(self):
        try:
            return self._setup_py
        except AttributeError:
            self._setup_py = None
            if self._get_metadata_value('install_setup', True):
                filename = self._get_metadata_value('setupscript', 'setup.py')
                rootdir = os.path.dirname(self.metafile)
                #rootdir = self._get_rootdir()
                filename = os.path.join(rootdir, filename)
                if os.path.isdir(filename):
                    filename = os.path.join(filename, 'setup.py')
                if os.path.exists(filename):
                    self._setup_py = filename
            return self._setup_py
```

Also, I'm thinking it might be best for the default "install_setup" to be `True`.  So, by default, the setup script is installed if it is there.  That seems like the more likely user expectation.

pyperformance/data-files/benchmarks/bm_ctypes/run_benchmark.py
```diff
@@ -0,0 +1,104 @@
...
+        ctypes.c_int,
+        ctypes.c_int,
+    ]
+    void_foo_int_int_int_int.restype = None
+
+    void_foo_constchar = ext.void_foo_constchar
+    void_foo_constchar.argtypes = [ctypes.c_char_p]
+    void_foo_constchar.restype = None
+
+    return benchmark(loops)
```
ericsnowcurrently, line 51: nice re-use! 😄

doc/custom_benchmarks.rst
```diff
@@ -333,6 +333,7 @@ tool.extra_opts [str]     X
 tool.inherits   file           
 tool.runscript  file      X    
 tool.datadir    file      X    
+tool.setup_py   bool
```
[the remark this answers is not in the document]
ericsnowcurrently: If adding "setupscript" seems premature, we can skip it.  (I'm just trying to keep custom benchmarks in mind.  Any friction in writing a benchmark can mean the difference between someone doing it or not.)

ericsnowcurrently (changes requested):
Thanks for the changes.  I just have a couple more suggestions.

pyperformance/venv.py
```diff
@@ -24,6 +24,8 @@ def from_benchmarks(cls, benchmarks):
         for bench in benchmarks or ():
             filename = bench.requirements_lockfile
             self._add_from_file(filename)
+            if bench.setup_py:
+                self._add(os.path.dirname(bench.metafile))
```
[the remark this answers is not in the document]
mdboom: I'm not sure I get this one.  `setup_py` is not a boolean to indicate there is one, and then you pass the directory it contains to `pip` (not the file itself).  Maybe renaming this flag to `needs_build` would be more accurate (since that how `pip` see it -- this directory has something that needs to be built, and `setup.py` is just an implementation detail...)

doc/custom_benchmarks.rst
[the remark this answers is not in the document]
mdboom: Yeah -- in addition, I don't think `pip` makes it possible.  You don't tell it "here's a setup.py", you tell it "here's a source directory to install, and it knows that `setup.py` is one of the ways to build it..."

pyperformance/_benchmark.py
mdboom: See above for why the filename may not be necessary.

But no harm in changing the default to `True` (I think).

pyperformance/venv.py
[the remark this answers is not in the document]
ericsnowcurrently: Sorry for the confusion.  My point is that we should record the directory name on the Benchmark object rather than using `os.path.dirname()` here.  We could call it something like `setup_py_dir` if you prefer.

doc/benchmarks.rst
```diff
@@ -109,6 +109,16 @@ the AES block cipher algorithm and the common modes of operation (CBC, CFB,
 CTR, ECB and OFB).
 
 
+ctypes
+------
+
+Benchmark to measure the function call overhead of calling C functions using ctypes.
+
+The `ctypes` benchmark lets `ctypes` infer the argument types from the passed in
```
kumaraditya303, line 117:
```suggestion
The ``ctypes`` benchmark lets ``ctypes`` infer the argument types from the passed in
```

doc/benchmarks.rst
```diff
@@ -109,6 +109,16 @@ the AES block cipher algorithm and the common modes of operation (CBC, CFB,
...
+values. The `ctypes_argtypes` benchmark `explicitly specifies the argument types
```
kumaraditya303, line 118:
```suggestion
values. The ``ctypes_argtypes`` benchmark `explicitly specifies the argument types
```

pyperformance/data-files/benchmarks/bm_ctypes/run_benchmark.py
```diff
@@ -0,0 +1,108 @@
...
+    void_foo_int = ext.void_foo_int
+    void_foo_int_int = ext.void_foo_int_int
+    void_foo_int_int_int = ext.void_foo_int_int_int
+    void_foo_int_int_int_int = ext.void_foo_int_int_int_int
+    void_foo_constchar = ext.void_foo_constchar
+
+    # Test calling the functions using the implied arguments mechanism
+    t0 = pyperf.perf_counter()
+
+    for i in range(loops):
```
kumaraditya303, line 66: Take the iterator of loop to reduce noise.

pyperformance/data-files/benchmarks/bm_ctypes/setup.py
```diff
@@ -0,0 +1,11 @@
+from distutils.core import setup, Extension
```
kumaraditya303, line 1: `distutils` is deprecated, you can use `setuptools`

kumaraditya303:
Seems like you are first time contributor hence GitHub is not running the CI. Someone with write access would have to manually trigger it :(.

pyperformance/data-files/benchmarks/bm_ctypes/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[project]
+name = "pyperformance_bm_ctypes"
+requires-python = ">=3.8"
+dependencies = ["pyperf", "setuptools"]
+urls = {repository = "https://github.com/python/pyperformance"}
+dynamic = ["version"]
+
+[tool.pyperformance]
+name = "ctypes"
+tags = "extension"
```
kumaraditya303, line 10:
```suggestion
tags = "extension"

```
File missing trailing whitespace.

kumaraditya303:
LGTM, but I am a bit concerned that benchmarking all of the call argtypes in one iteration can hide regressions, but I can't think of anyway to fix that apart from splitting this benchmark  further to each call argtypes.

kumaraditya303 (approved):

pyperformance/data-files/benchmarks/bm_ctypes/cmodule.c
```diff
@@ -0,0 +1,54 @@
+#include <Python.h>
+
+#if defined(_WIN32) || defined(__CYGWIN__)
+#define EXPORTED_SYMBOL __declspec(dllexport)
+#else
+#define EXPORTED_SYMBOL
```
kumaraditya303, line 6: Maybe export the symbol on other platforms too? 
The following should work for gcc and clang:
```suggestion
#define EXPORTED_SYMBOL __attribute__((visibility ("default")))
```

kumaraditya303: You can also use the one which Python uses: `Py_EXPORTED_SYMBOL` and avoid this altogether.

mdboom: Unfortunately, `Py_EXPORTED_SYMBOL` doesn't exist in Python 3.8.  And the explicit visibility isn't necessary on gcc or clang where visible is the default.

I could copy the full implementation of `Py_EXPORTED_SYMBOL` in CPython here and use it, though, if that's just "less surprising".

kumaraditya303: > Py_EXPORTED_SYMBOL doesn't exist in Python 3.8. 

Oh, I didn't know that, probably best left as it is then.

pyperformance/_benchmark.py
```diff
@@ -164,6 +164,10 @@ def runscript(self):
     def extra_opts(self):
         return self._get_metadata_value('extra_opts', ())
 
+    @property
+    def install_setup(self):
+        return self._get_metadata_value('install_setup', False)
```
ericsnowcurrently, line 169: Looking back on [my previous review](https://github.com/python/pyperformance/pull/197#pullrequestreview-985170156), I still think it would be better to provide the path to the setup.py file, if there is one, or `None` otherwise.

```suggestion
    @property
    def setup_py(self):
        if not self._get_metadata_value('install_setup', False):
            return None
        filename = os.path.join(os.path.dirname(self.metafile), 'setup.py'))
        return filename if os.path.exists(filename) else None
```

The fact that pip only supports passing the directory isn't something we should design around here.  We can handle that quirk in the code where we interact with pip.

pyperformance/venv.py
```diff
@@ -24,6 +24,8 @@ def from_benchmarks(cls, benchmarks):
         for bench in benchmarks or ():
             filename = bench.requirements_lockfile
             self._add_from_file(filename)
+            if bench.install_setup:
+                self._add(os.path.dirname(bench.metafile))
```
ericsnowcurrently, line 28: Along with my other comment:

```suggestion
            if bench.setup_py:
                # pip doesn't support installing a setup.py,
                # but it does support installing from the directory it is in.
                self._add(os.path.dirname(bench.setup_py))
```

pyperformance/data-files/benchmarks/bm_ctypes/run_benchmark.py
```diff
@@ -0,0 +1,110 @@
...
+    return pyperf.perf_counter() - t0
+
+
+def add_cmdline_args(cmd, args):
+    if args.argtypes:
+        cmd.append("--argtypes")
+
+
+if __name__ == "__main__":
+    runner = pyperf.Runner(add_cmdline_args=add_cmdline_args)
```
ericsnowcurrently, line 101: \<rant>
"add_cmdline_args" confuses me every time.  The parameter should be named something like "add_worker_argv".  It would be even better if there were also an "inherit_worker_argv" arg (boolean) so the callback is unnecessary.
\</rant>

Anyway, nothing to do here.

pyperformance/data-files/benchmarks/bm_ctypes/run_benchmark.py
```diff
@@ -0,0 +1,110 @@
...
+    runner.metadata["description"] = "ctypes function call overhead benchmark"
+
+    runner.argparser.add_argument("--argtypes", action="store_true")
+    options = runner.parse_args()
+
+    if options.argtypes:
+        runner.bench_time_func("ctypes_argtypes", benchmark_argtypes, inner_loops=3)
+    else:
+        runner.bench_time_func("ctypes", benchmark, inner_loops=3)
```
ericsnowcurrently, line 110: Out of curiosity, why 3 inner loops?  (Why not more?)

pyperformance/data-files/benchmarks/bm_ctypes/run_benchmark.py
```diff
@@ -0,0 +1,110 @@
...
+        void_foo_int_int_int_int(1, 2, 3, 4)
+        void_foo_constchar(b"bytes")
+
+        void_foo_void()
+        int_foo_int(1)
+        void_foo_int(1)
+        void_foo_int_int(1, 2)
+        void_foo_int_int_int(1, 2, 3)
+        void_foo_int_int_int_int(1, 2, 3, 4)
+
```
ericsnowcurrently, line 91: Did you intentionally leave `void_foo_constchar(b"bytes")` off this group of calls?  If so, a short comment on why would be helpful here.

pyperformance/data-files/benchmarks/bm_ctypes/run_benchmark.py
```diff
@@ -0,0 +1,110 @@
...
+    t0 = pyperf.perf_counter()
+
+    for _ in range_it:
+        void_foo_void()
+        int_foo_int(1)
+        void_foo_int(1)
+        void_foo_int_int(1, 2)
+        void_foo_int_int_int(1, 2, 3)
+        void_foo_int_int_int_int(1, 2, 3, 4)
+        void_foo_constchar(b"bytes")
```
ericsnowcurrently, line 75: Why manually unroll the loop 3 times?  Couldn't you do `range(loops*3)` instead?  If there's a reason then it might help to have a short comment to explain.

ericsnowcurrently (changes requested):
Mostly LGTM.  I've left a few comments on a few minor things and to get clarification in a couple spots.

pyperformance/venv.py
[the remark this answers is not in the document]
ericsnowcurrently: I've revisited this in https://github.com/python/pyperformance/pull/197#discussion_r930154904.

pyperformance/_benchmark.py
ericsnowcurrently: see https://github.com/python/pyperformance/pull/197#discussion_r930154904

pyperformance/_benchmark.py
mdboom: You're suggesting we just pass `dirname(filename)` to pip?  That would work.

pyperformance/data-files/benchmarks/bm_ctypes/run_benchmark.py
mdboom: No, that's just an oversight.

pyperformance/data-files/benchmarks/bm_ctypes/run_benchmark.py
mdboom: I think this is just what I copied from another example.  I agree, this is a "fast" benchmark, it should have more repetitions.

pyperformance/data-files/benchmarks/bm_ctypes/run_benchmark.py
mdboom: Someone somewhere had suggested it, but I can't find it now.  It's certainly simpler to not unroll, so I think I'll just remove the repetition now.

pyperformance/venv.py
```diff
@@ -24,6 +24,10 @@ def from_benchmarks(cls, benchmarks):
         for bench in benchmarks or ():
             filename = bench.requirements_lockfile
             self._add_from_file(filename)
+            if bench.setup_py:
+                # pip doesn't support installing a setup.py,
+                # but it does support installing from the directory it is in.
+                self._add(os.path.dirname(bench.setup_py))
```
ericsnowcurrently, line 30: It might be better to handle this in `_pip.install_requirements()`, since it is a pip quirk, but I'm not going to stress over it.

```suggestion
                self._add(bench.setup_py)
```

(See the following comment for the `_pip.install_requirements()` changes.)

Again, I'm not going to block merging on this, though you're welcome to take my suggestion. 🙂

pyperformance/_pip.py
```diff
@@ -150,7 +150,7 @@ def install_requirements(reqs, *extra,
     if upgrade:
         args.append('-U')  # --upgrade
     for reqs in [reqs, *extra]:
-        if os.path.exists(reqs):
+        if os.path.isfile(reqs):
             args.append('-r')  # --requirement
         args.append(reqs)
```
ericsnowcurrently, line 155: As suggested above, the fact that you can't pass the full setup.py filename to pip should probably be handled here.  So something like:

```python
    for req in [reqs, *extra]:
        if os.path.isfile(req):
            name = os.path.basename(req)
            if re.match(r'^setup.*\.py$', name):
                if name != setup.py:
                    raise ValueError(f'pip doesn't support installing from {req!r}')
                req = os.path.dirname(req)
            else:
                args.append('-r')  # --requirement
        args.append(req)
```

Again, I'm not going to block merging on this.

ericsnowcurrently (commented):
A quick follow-up suggestion...

pyperformance/_pip.py
mdboom: Yeah -- I agree it makes sense to move the handling of this to here.  I'm not sure I follow the regex suggestion, though.  Are there cases where the filename wouldn't literally be "setup.py"?

mdboom: In any case, I moved this logic here, as you suggested, but just kept it to a direct string comparison to `setup.py`.  Let me know if the regex makes sense, otherwise, I think this is good to go.

ericsnowcurrently: Yeah, the regex is overkill for a less-than-likely hypothetical.  If we need to support more than setup.py, we can do that later.

So:

```python
    for req in [reqs, *extra]:
        if os.path.isfile(req):
            if os.path.basename(req) == 'setup.py':
                req = os.path.dirname(req)
            else:
                args.append('-r')  # --requirement
        args.append(req)
```

pyperformance/_pip.py
```diff
@@ -149,10 +149,16 @@ def install_requirements(reqs, *extra,
...
         args.append('-U')  # --upgrade
-    for reqs in [reqs, *extra]:
-        if os.path.exists(reqs):
-            args.append('-r')  # --requirement
-        args.append(reqs)
+    for req in [reqs, *extra]:
+        if os.path.isfile(req):
+            name = os.path.basename(req)
+            if name == "setup.py":
+                req = os.path.dirname(req)
```
ericsnowcurrently, line 156: The reason we use the dirname isn't obvious, so it may be worth adding a comment here indicating pip's limitations.

pyperformance/venv.py
```diff
@@ -24,6 +24,10 @@ def from_benchmarks(cls, benchmarks):
         for bench in benchmarks or ():
             filename = bench.requirements_lockfile
             self._add_from_file(filename)
+            if bench.setup_py:
+                # pip doesn't support installing a setup.py,
+                # but it does support installing from the directory it is in.
```
ericsnowcurrently, line 29: This comment is what I was thinking of above.  Consider moving it there.

ericsnowcurrently (approved):
LGTM

There's only one small suggestion, which I'll leave to your discretion.

PR #200, opened 2022-05-14
Add tomli_loads benchmark

kumaraditya303:


kumaraditya303:
It seems like `requires-python = ">=3.11"` has no effect and seems broken.

pyperformance/data-files/benchmarks/bm_tomllib_loads/run_benchmark.py
```diff
@@ -0,0 +1,28 @@
+"""
+Benchmark ``loads()`` function of the ``tomllib`` stdlib module
+on a large TOML file of GitHub's real world data.
```
ericsnowcurrently, line 3: Please identify how the data file was generated.  It is useful to be able to reproduce it (or something like it).

Also, is there any PII (personally-identifiable information) in the data that isn't yours or that you don't want exposed?

ericsnowcurrently (changes requested):
Thanks for working on this!  It mostly LGTM.  There's only one thing I'd like added.

pyperformance/data-files/benchmarks/bm_tomllib_loads/run_benchmark.py
kumaraditya303: The data was generated by this script https://gist.github.com/kumaraditya303/9b6f76f29e3ac1962320df0c7f9a685a.
and is public issues data of the cpython repo converted to toml.

kumaraditya303: Added link to script.

ericsnowcurrently: Why not include that code as a script in the benchmark directory?

kumaraditya303: I thought that it would be unnecessary as that script is needed to be run once. However thinking now, I would improve that script to output json and toml in a seprate PR so that the json benchmark can also use real world data instead of dummy data. What do you think?

ericsnowcurrently: That's good, but it would make sense to start off with the script from the gist as part of this PR.  Even if it's only run once, having the code for that next to the benchmark is still helpful.

kumaraditya303: Okay, will add tomorrow.

kumaraditya303: Added

pyperformance/data-files/benchmarks/bm_tomllib_loads/generate_data.py
```diff
@@ -0,0 +1,22 @@
...
+    page = 1
+    all_issues = []
+    while page <= 10:
+        with urlopen(f"{BASE_URL}&page={page}") as response:
+            issues = json.loads(response.read())
+            if not issues:
+                break
+            all_issues.extend(issues)
+            print(f"Page: {page} Total Issues: {len(all_issues)}")
+        page += 1
```
ericsnowcurrently, line 17: FWIW, this is often spelled as:
```suggestion
    all_issues = []
    for page in range(1, 11):
        with urlopen(f"{BASE_URL}&page={page}") as response:
            issues = json.loads(response.read())
            if not issues:
                break
            all_issues.extend(issues)
            print(f"Page: {page} Total Issues: {len(all_issues)}")
```

ericsnowcurrently (approved):
LGTM

Thanks for working on this.

pyperformance/data-files/benchmarks/bm_tomllib_loads/generate_data.py
kumaraditya303: done

ericsnowcurrently:
We can merge this once the tests are passing.  Do you know what the problem is?

ericsnowcurrently:
Ah, right.  I'll take a look at that real quick and then merge it.

mdboom:
Maybe we should have this install `toml` from PyPI instead of only using the stdlib one?  That would give us more Python version coverage.

hauntsaninja:
(tomllib is based on pypi's tomli, not toml)

PR #201, opened 2022-05-15
Add benchmark for deepcopy

### LINKED ISSUE #199: Benchmark suggestion: deepcopy
One python method that is currently not in the pyperformance benchmarks is `deepcopy` (at least not in the benchmarks that execute on my system).

Can we add a benchmark specifically for `deepcopy`?

In [#91610](https://github.com/python/cpython/pull/91610) (reimplement `deepcopy` in C) there is a simple benchmarking script for `deepcopy`. It is a simple execution of `deepcopy` on a `dict` and a `dataclass`. The benchmark is really simple (not embedded program), but it is representative for the cost of `deepcopy` that I encountered in larger projects. 

If required, I can create a PR with a benchmark similar to the one in [#91610](https://github.com/python/cpython/pull/91610).

eendebakpt:
One python method that is currently not in the pyperformance benchmarks is `deepcopy`. This PRs adds a benchmark that tests performance of `deepcopy` on a nested dictionary and a dataclass.

The dictionary and dataclass have been chosen as example of structures can are often copied and then modified. (e.g. passing configuration settings to a method that modifies one of the settings before passing it on)

Fixes #199

### DIFF
--- a/pyperformance/data-files/benchmarks/bm_deepcopy/run_benchmark.py
+++ b/pyperformance/data-files/benchmarks/bm_deepcopy/run_benchmark.py
@@ -18,20 +18,17 @@ class A:
     
 
 def benchmark(n):
-    a={'list': [1,2,3,43], 't': (1,2,3), 'str': 'hello', 'subdict': {'a': True}}
-    dc=A('hello', [1,2,3], True)
-    
+    a = {'list': [1,2,3,43], 't': (1,2,3), 'str': 'hello', 'subdict': {'a': True}}
+    dc = A('hello', [1,2,3], True)
     for ii in range(n):
-        
         for jj in range(60):
             _ = copy.deepcopy(a)
-            
         for s in ['red', 'blue', 'green']:
             dc.string = s
             for ii in range(10):
                 dc.lst[0] = ii
                 for b in [True, False]:
-                    dc.boolean=b
+                    dc.boolean = b
                     _ = copy.deepcopy(dc)
                 
 
--- a/pyperformance/data-files/benchmarks/bm_deepcopy/run_benchmark.py
+++ b/pyperformance/data-files/benchmarks/bm_deepcopy/run_benchmark.py
@@ -1,0 +1,42 @@
+"""
+Benchmark to measure performance of the python builtin method copy.deepcopy
+
+Performance is tested on a nested dictionary and a dataclass
+
+Author: Pieter Eendebak
+
+"""
+import copy
+import pyperf
+from dataclasses import dataclass
+
+@dataclass
+class A:
+    string : str
+    lst : list
+    boolean : bool
+    
+
+def benchmark(n):
+    a={'list': [1,2,3,43], 't': (1,2,3), 'str': 'hello', 'subdict': {'a': True}}
+    dc=A('hello', [1,2,3], True)
+    
+    for ii in range(n):
+        
+        for jj in range(60):
+            _ = copy.deepcopy(a)
+            
+        for s in ['red', 'blue', 'green']:
+            dc.string = s
+            for ii in range(10):
+                dc.lst[0] = ii
+                for b in [True, False]:
+                    dc.boolean=b
+                    _ = copy.deepcopy(dc)
+                
+
+if __name__ == "__main__":
+    runner = pyperf.Runner()
+    runner.metadata['description'] = "deepcopy benchmark"
+
+    runner.bench_func('deepcopy', benchmark, 200)
--- a/pyperformance/data-files/benchmarks/bm_deepcopy/run_benchmark.py
+++ b/pyperformance/data-files/benchmarks/bm_deepcopy/run_benchmark.py
@@ -61,18 +61,21 @@ def benchmark(n):
     }
     dc = A('hello', [1, 2, 3], True)
 
-    t0 = pyperf.perf_counter()
+    dt = 0
     for ii in range(n):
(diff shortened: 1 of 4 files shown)


pyperformance/data-files/benchmarks/bm_deepcopy/run_benchmark.py
```diff
@@ -0,0 +1,42 @@
...
+                for b in [True, False]:
+                    dc.boolean=b
+                    _ = copy.deepcopy(dc)
+                
+
+if __name__ == "__main__":
+    runner = pyperf.Runner()
+    runner.metadata['description'] = "deepcopy benchmark"
+
+    runner.bench_func('deepcopy', benchmark, 200)
```
ericsnowcurrently, line 42: Consider using `runner.bench_time_func()`.  It is better suited for micro benchmarks like this.  I expect the `copy.deepcopy()` call is not massively slower that all the other stuff going on in `benchmark()`.  That means all of that may skew the results.  Using bench_time_func() helps avoid that.  There are several examples in the other benchmarks.

pyperformance/data-files/benchmarks/bm_deepcopy/run_benchmark.py
```diff
@@ -0,0 +1,42 @@
...
+@dataclass
+class A:
+    string : str
+    lst : list
+    boolean : bool
+    
+
+def benchmark(n):
+    a={'list': [1,2,3,43], 't': (1,2,3), 'str': 'hello', 'subdict': {'a': True}}
+    dc=A('hello', [1,2,3], True)
```
ericsnowcurrently, line 22: If the objective of this micro benchmark is to monitor for `copy.deepcopy()` performance regressions, then we need to be sure the benchmark has sufficient code coverage.  (It may even be worth running coverage.py on it.)  At the least consider deepcopying other supported objects, including objects that implement different parts of the copy protocol (`__reduce__()`, `__setstate__()`, etc.).

ericsnowcurrently (changes requested):
This looks mostly good.  Aside from a few linting issues, there are only a couple small concerns.

pyperformance/data-files/benchmarks/bm_deepcopy/run_benchmark.py
```diff
@@ -0,0 +1,42 @@
...
+
+@dataclass
+class A:
+    string : str
+    lst : list
+    boolean : bool
+    
+
+def benchmark(n):
+    a={'list': [1,2,3,43], 't': (1,2,3), 'str': 'hello', 'subdict': {'a': True}}
```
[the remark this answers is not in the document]
ericsnowcurrently: No need to apologize.  You could use any tool you like.  The project doesn't use one.  Personally, I don't bother with tools.  It doesn't have to be perfect.  It's just readability that counts.  When things are crammed together they're harder to read.

pyperformance/data-files/benchmarks/bm_deepcopy/run_benchmark.py
eendebakpt: I added 2 more benchmarks for the `__reduce__/__setstate__` and `memo` functionality in `deepcopy`. @ericsnowcurrently

eendebakpt:
> This looks mostly good. Aside from a few linting issues, there are only a couple small concerns.

@ericsnowcurrently Changes made.

ericsnowcurrently:
Thanks, @eendebakpt.  I'll take a look as soon as I can.

pyperformance/data-files/benchmarks/bm_deepcopy/run_benchmark.py
```diff
@@ -0,0 +1,85 @@
...
+        for jj in range(30):
+            _ = copy.deepcopy(a)
+        for s in ['red', 'blue', 'green']:
+            dc.string = s
+            for kk in range(5):
+                dc.lst[0] = kk
+                for b in [True, False]:
+                    dc.boolean = b
+                    _ = copy.deepcopy(dc)
+    dt = pyperf.perf_counter() - t0
```
ericsnowcurrently, line 75: Please make sure the loops, etc. don't add a significant overhead.  If they do, consider tracking only the deepcopy calls:

```suggestion
    dt = 0
    for ii in range(n):
        for jj in range(30):
            t0 = pyperf.perf_counter()
            _ = copy.deepcopy(a)
            dt += pyperf.perf_counter() - t0
        for s in ['red', 'blue', 'green']:
            dc.string = s
            for kk in range(5):
                dc.lst[0] = kk
                for b in [True, False]:
                    dc.boolean = b
                    t0 = pyperf.perf_counter()
                    _ = copy.deepcopy(dc)
                    dt += pyperf.perf_counter() - t0
```

ericsnowcurrently (changes requested):
Mostly LGTM.  There are a few really minor things that I won't even bother mentioning. 😄  The only thing worth looking at is the overhead of everything going on the in the various loops in the one benchmark.  My recommendation doesn't impact readability all that much so you might as well apply it, just in case.

pyperformance/data-files/benchmarks/bm_deepcopy/run_benchmark.py
ericsnowcurrently: Other than that, this should be good to go.

ericsnowcurrently (approved):
LGTM

PR #206, opened 2022-05-22
Add generators benchmark

kumaraditya303:


gvanrossum (commented):
The code is fine, but it looks more like a microbenchmark. Have you compared 3.10 and 3.11 yet with this?

@ericsnowcurrently Do you think this makes a reasonable addition? I'm not sure where we stand on adding more microbenchmarks.

@kumaraditya303 The new files are missing trailing newlines.

pyperformance/data-files/benchmarks/bm_generators/run_benchmark.py
```diff
@@ -0,0 +1,48 @@
...
+
+from __future__ import annotations
+
+from collections.abc import Iterator
+
+import pyperf
+
+
+class Tree:
+    def __init__(self, value: int, left: Tree | None, right: Tree | None) -> None:
```
gvanrossum, line 16: FWIW maybe `value` should be the second parameter, since the iterator iterates in in-order? The values would then be provided in the same order in which they are produced.

gvanrossum (commented):
One minor suggestion.

gvanrossum:
@kumaraditya303 We discussed adding microbenchmarks and the conclusion is that it's okay. So can you address the minor nits? Also, I am concerned about the greenlet-related failure in the 3.11-dev test. I am curious though -- how old is the 3.11-dev version on Ubuntu? Maybe it's an older alpha?

pyperformance/data-files/benchmarks/bm_generators/run_benchmark.py
kumaraditya303: Done

kumaraditya303:
3.11 is failing because of compilation error when compiling greenlet. See the tracking issue https://github.com/python/cpython/issues/92898, it is fixed in main branch and 3.11 but no release has that fix yet.

kumaraditya303:
> We discussed adding microbenchmarks and the conclusion is that it's okay. 

One thing microbenchmarks are very good at is finding regressions as it focused on one thing only and can easily identify regressions.

gvanrossum (approved):

gvanrossum:
Thanks for the benchmark, and thanks for the link to the issue with greenlet!

kumaraditya303:
@gvanrossum Benchmark results comparing 3.10 release and 3.12 main branch:

| Benchmark  | 3.10   | 3.12                 |
|------------|:------:|:--------------------:|
| generators | 299 ms | 206 ms: 1.46x faster |

gvanrossum:
Nice!

PR #211, opened 2022-05-26
use deadsnakes for nightly 3.11 CI 

kumaraditya303:


kumaraditya303:
Finally CI is green and Python 3.11 is passing the tests.

.github/workflows/main.yml
```diff
@@ -30,14 +30,17 @@ jobs:
...
-            python: "3.9"
+            python: "3.10"
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python }}
       uses: actions/setup-python@v2
+      if: "!endsWith(matrix.python, '-dev')"
+      with:
+        python-version: ${{ matrix.python }}
+    - uses: deadsnakes/action@v2.1.1
```
gvanrossum, line 42: Shouldn’t this one also have a name field?

kumaraditya303: It does not matter as the CI title already shows the Python version.

gvanrossum, line 42:
```suggestion
    - name: Set up Python ${{ matrix.python }} using deadsnakes
    - uses: deadsnakes/action@v2.1.1
```

gvanrossum (approved):

gvanrossum:
Well, I felt it was important to be consistent with (most) other entries, so I added the name. Thanks for fixing this!

gvanrossum:
Something feels off though. I got some emails from the CI that no tests were run.

kumaraditya303:
Yes, seems like CI didn't run on main. GitHub has been crashing for the last two days.

kumaraditya303:
https://github.com/python/pyperformance/pull/211/commits/8741703dc7e029378440ae346186bf6f7ab01a90 broke the CI as it uses invalid yaml syntax. I created #212 to fix it.

PR #212, opened 2022-05-27
fix yaml syntax to fix CI

kumaraditya303:


gvanrossum:
Yeah, sorry about that! I should have waited for you.

PR #213, opened 2022-06-06
Add coverage benchmark

kumaraditya303:
Add a coverage benchmark to help benchmarking and avoid regressions such as https://github.com/python/cpython/issues/93516.


gvanrossum (commented):
LGTM, thanks for doing something about this.

Question for @kumaraditya303 -- Have you actually tried this before and after the regression you mentioned as example?

Question for @ericsnowcurrently -- Do you agree this is useful?

kumaraditya303:
> Have you actually tried this before and after the regression you mentioned as example?

@gvanrossum: The regression I linked above [^1] is not caused by a single change but rather accumulation of multiple slowdown caused by multiple changes of which the most significant ones being https://github.com/python/cpython/pull/91666 and https://github.com/python/cpython/pull/27077. Since there is no benchmark in pyperformance to exercise coverage performance, the slowdown was unnoticed. Once this benchmark is added, it will help us pinpoint which commits caused the most slowdown with a simple `git bisect` between `3.10` and `main` branches.

[^1]: https://github.com/python/pyperformance/pull/213#issue-1261503016

gvanrossum:
Nevertheless it would be interesting to see the numbers of running this BM on the state before and after said commits to demonstrate its usefulness.

kumaraditya303:
I am running the benchmarks now, will post soon.

kumaraditya303:
@gvanrossum  Benchmarking before and after https://github.com/python/cpython/commit/944fffee8916cb94321fa33cd3a43f4108717746 commit

Results:

| Benchmark | before | after                |
|-----------|:------:|:--------------------:|
| coverage  | 161 ms | 213 ms: 1.32x slower |

ericsnowcurrently:
> Question for @ericsnowcurrently -- Do you agree this is useful?

Yeah, it's a helpful indicator of impact.

PR #215, opened 2022-06-13
Allow for specifying local wheels and sdists as dependencies

mdboom:
It is useful, especially when creating custom versions of dependencies that are compatible with CPython main, to specify local file paths as dependencies.  Unfortunately, that's blocked by a feature that treats any dependency that matches a file path as a requirements file.  This makes that check a little more specific which allows for installing wheels.

This feels like the "safe" backward compatible change, but to be honest, I don't know why this feature exists in the first place -- it doesn't seem to be documented and the unit tests don't fail when you remove this whole `if` clause.  I would also be ok with just removing this `if` clause if preferred.


kumaraditya303:
FWIW, when I create a custom patched version of a package I install it with pip directly using my own fork and it works fine.

kumaraditya303 (approved):
LGTM, Could you add a test for this? I think a simple test with mock would suffice. Thanks!

pyperformance/tests/data/MANIFEST
```diff
@@ -0,0 +1,4 @@
+[benchmarks]
+
+name	metafile
+local_wheel	<local>
```
arhadthedev, line 4: Is missing end-of-line intentional here?

kumaraditya303 (approved):

pyperformance/venv.py
```diff
@@ -232,7 +232,7 @@ def _install_pyperf_optional_dependencies(self):
                 print("WARNING: failed to install %s" % req)
                 pass
 
-    def ensure_reqs(self, requirements=None):
+    def ensure_reqs(self, requirements=None, cwd=None):
```
ericsnowcurrently, line 235: Instead of passing it in, we can infer `cwd` from `requirements` in the `elif hasattr(requirements, 'requirements_lockfile'):` case.  That seems appropriate for `VenvForBenchmarks`, which has the necessary context.

pyperformance/run.py
```diff
@@ -94,7 +94,7 @@ def run_benchmarks(should_run, python, options):
         )
         try:
             # XXX Do not override when there is a requirements collision.
-            venv.ensure_reqs(bench)
+            venv.ensure_reqs(bench, cwd=bench._rootdir)
```
ericsnowcurrently, line 97: As noted, passing `cwd=bench._rootdir` seems a bit redundant.

pyperformance/_venv.py
```diff
@@ -239,13 +239,14 @@ def upgrade_pip(self, *, installer=True):
...
 
-    def ensure_reqs(self, *reqs, upgrade=True):
+    def ensure_reqs(self, *reqs, upgrade=True, cwd=None):
         print("Installing requirements into the virtual environment %s" % self.root)
         ec, _, _ = _pip.install_requirements(
             *reqs,
             python=self.python,
             env=self._env,
             upgrade=upgrade,
+            cwd=cwd
```
ericsnowcurrently, line 249: Presumably this is so pip will find the wheel file.  Something feels off to using cwd for this.

What are the alternatives?  What about:

* pass the resolved path to the wheel file as a requirement?
* use pip's `--index-url`, `--extra-index-url`, or `--find-links` args?

ericsnowcurrently (changes requested):
For the most part this seems good.  There is basically just one thing that's throwing me off.

pyperformance/_venv.py
mdboom: Yeah -- `--find-links` seems like the better solution here. Let me poke at this and see if I can get it to work.

ericsnowcurrently (approved):
LGTM

mdboom:
Had to go back to the drawing board on this approach, since `install_requirements` doesn't actually have any context to know the directory of the requirements.txt file.  I instead moved the resolving to a full path to `_add_from_file` which parses the `requirements.txt` file.

ericsnowcurrently (approved):
LGTM

(and thanks for adding good tests)

PR #219, opened 2022-06-22
Add `mdp` back into the `default` group

brandtbucher:
When the `mdp` benchmark was [added](https://github.com/python/pyperformance/pull/14) over 5 years ago, it was disabled by default because it took too long to run (reportedly 7-8 seconds).

However, this situation has improved somewhat (I time about 3 seconds on recent Python versions), and we're currently trying to move towards including more long-running benchmarks in the default suite.

It's a good benchmark, so let's enable it again.

### DIFF
--- a/pyperformance/data-files/benchmarks/MANIFEST
+++ b/pyperformance/data-files/benchmarks/MANIFEST
@@ -73,4 +73,3 @@ xml_etree	<local>
 
 
 [group default]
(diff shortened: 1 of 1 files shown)


mdboom (approved):
I'm not an official reviewer here, but +1 from me.  This seems aligned with recent discussions about types of benchmarks.

PR #224, opened 2022-08-05
Fix bm_raytrace filename option

corona10:
`array.tostring` was replaced with `array.tobytes` by Python 3.2.
`array.tostring` does not work anymore and this feature was broken for many years :(
ref: https://docs.python.org/3/library/array.html#array.array.tobytes


ericsnowcurrently (approved):
LGTM

PR #231, opened 2022-08-18
Fix #214: Always upload, even when some benchmarks fail

### LINKED ISSUE #214: Ensure results are uploaded even if a single benchmark runs fail
Unfortunately, even if a single benchmark fails right now, we don't upload the results to speed.python.org (and maybe we don't even write the result files). We should be a bit more flexible and at least upload the benchmarks that have succeeded in the run.

mdboom:
Currently, if a single benchmark fails, none of the results are uploaded for that entire `compile` command run.  With this change, we always upload the results if only individual benchmarks fail.  If the compilation fails or "global" venv building fails, of course, we still don't upload.

This changes the reporting for `compile_all`: Since uploading now almost always happens, there's no reason to report it as a special case.  Instead, we note when it fails as a result of a benchmark failing, and report that some benchmarks didn't work for that run.

Testing this is really tricky, given that we don't have any tests at all that capture and check the output of `compile` or `compile_all` currently.  I'm not sure it's worth the effort to clean that up as part of this specific PR.

I did manually check the following cases though:

- The benchmark throws an exception
- The benchmark dies (with `os._exit(1)`)
- The dependencies for a benchmark can't be installed

Any other important use cases to test?

### DIFF
--- a/pyperformance/compile.py
+++ b/pyperformance/compile.py
@@ -703,7 +703,7 @@ class BenchmarkRevision(Application):
             python = None
 
         failed = self.run_benchmark(python)
-        if not failed and self.conf.upload:
+        if self.conf.upload:
             self.upload()
         return failed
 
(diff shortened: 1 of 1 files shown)


pablogsal:
I think this covers all cases we care about. If we discover more failure modes we can fix them afterwards.

Thanks for the fix @mdboom. Great work 🤘

PR #236, opened 2022-09-09
Add jobs field in compile section to specify make -j param

aisk:
For reduce compilation time cost in multi core machines.

### DIFF
--- a/pyperformance/compile.py
+++ b/pyperformance/compile.py
@@ -780,7 +780,7 @@ def parse_config(filename, command):
         except KeyError:
             return default
 
-    def getint(section, key, default):
+    def getint(section, key, default=None):
         return int(getstr(section, key, default))
 
     # [config]
--- a/pyperformance/compile.py
+++ b/pyperformance/compile.py
@@ -298,11 +298,13 @@ class Python(Task):
         configure = os.path.join(self.conf.repo_dir, 'configure')
         self.run(configure, *config_args)
 
+        argv = ['make']
         if self.conf.pgo:
             # FIXME: use taskset (isolated CPUs) for PGO?
-            self.run('make', 'profile-opt')
-        else:
-            self.run('make')
+            argv.append('profile-opt')
+        if self.conf.jobs:
+            argv.append('-j%d' % self.conf.jobs)
+        self.run(*argv)
 
     def install_python(self):
         program, _ = resolve_python(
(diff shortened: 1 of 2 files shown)


pyperformance/compile.py
```diff
@@ -300,9 +300,9 @@ def compile(self):
 
         if self.conf.pgo:
             # FIXME: use taskset (isolated CPUs) for PGO?
-            self.run('make', 'profile-opt')
+            self.run('make', 'profile-opt', '-j%d' % self.conf.jobs)
         else:
-            self.run('make')
+            self.run('make', '-j%d' % self.conf.jobs)
```
ericsnowcurrently, line 305: It may make sense to only add the flag if the config option is set:

```python
        argv = ['make']
        if self.conf.pgo:
            # FIXME: use taskset (isolated CPUs) for PGO?
            argv.append('profile-opt')
        if self.conf.jobs:
            argv.append('-j%d' % self.conf.jobs)
        self.run(*argv)
```

So the config default would be `None` instead of 1.

The reason to do this is: the current default could be something other than 1 given that there are other ways to specify the number of concurrent jobs to run.

ericsnowcurrently (changes requested):
I'm in favor of adding this and the change mostly looks good.  I have but one recommendation first.

pyperformance/compile.py
aisk: Should we apply the `-j` parameters with the pgo job?

ericsnowcurrently: It would be.  That's `if self.conf.jobs:` not `elif self.conf.jobs:`.  Perhaps I misunderstood your question?

aisk: Oh, I understood! Updated.

pyperformance/compile.py
```diff
@@ -778,6 +780,9 @@ def getboolean(section, key, default):
         except KeyError:
             return default
 
+    def getint(section, key, default):
+        return int(getstr(section, key, default))
```
ericsnowcurrently, line 784:
```suggestion
    def getint(section, key, default=None):
        return int(getstr(section, key, default))
```

ericsnowcurrently (changes requested):
Thanks for making that change.  There's just one more thing to fix: the default value.

ericsnowcurrently (approved):
LGTM

ericsnowcurrently:
Thanks for the patch, @aisk!

PR #240, opened 2022-11-04
Upgrade to pyperf 2.5.0

mdboom:
This, in particular, is to get the new pystats support in pyperf as part of:

  https://github.com/faster-cpython/tools/issues/115


mdboom:
See https://github.com/python-greenlet/greenlet/issues/323 for more info on the 3.12 breakage.

corona10 (approved):
lgtm :)

PR #243, opened 2022-11-07
Add benchmark to measure gc collection of a big chain of cycles

pablogsal:


pablogsal:
More to come, I will create a PR per benchmark

pyperformance/data-files/benchmarks/bm_gc_collect/run_benchmark.py
```diff
@@ -0,0 +1,64 @@
...
+CYCLES = 100
+LINKS = 20
+
+class Node:
+    def __init__(self):
+        self.next = None
+        self.prev = None
+    
+    def link_next(self, next):
+        self.next = next
```
markshannon, line 13: If you add `next.back = self` here you don't need the `link_back` method.

markshannon (commented):
OOI do you have any numbers as to what fraction of the runtime this spends in GC?

pablogsal:
> OOI do you have any numbers as to what fraction of the runtime this spends in GC?

By definition all of it, as the benchmark is only measuring `gc.collect()` call.

pyperformance/data-files/benchmarks/bm_gc_collect/run_benchmark.py
pablogsal: Done

ericsnowcurrently (approved):
LGTM

PR #244, opened 2022-11-07
Add a benchmark to measure gc traversal

pablogsal:


pyperformance/data-files/benchmarks/bm_gc_traversal/run_benchmark.py
```diff
@@ -0,0 +1,37 @@
...
+
+def benchamark_collection(loops, n_levels):
+    total_time = 0
+    all_cycles = create_recursive_containers(n_levels)
+    for _ in range(loops):
+        gc.collect()
+        # Main loop to measure
+        t0 = pyperf.perf_counter()
+        collected = gc.collect()
+        total_time += pyperf.perf_counter() - t0
```
ericsnowcurrently, line 27: Just to be sure I understand, the GC objects aren't cleaned up during `gc.collect()` because `all_cycles` is still bound in the locals.  That's why we can re-use it in each loop.  Is that right?

ericsnowcurrently (approved):
LGTM

There's one thing to clarify, but I think I've understood it correctly.  I'm approving the PR under that assumption. 🙂

pyperformance/data-files/benchmarks/bm_gc_traversal/run_benchmark.py
pablogsal: Yeah, this test is just benchmarking the gc traversing all those objects again and again so no need to remove them

nanjekyejoannah:
Why dont you create one benchmark as we know its supposed to be a workload, then things like traversal, collection are just metrics? The PyPy folks have a workload called gcbench we can reference. I stand to be corrected but I am not familiar with folks writing a specific workload for every metric they want to measure. Rather its applying some metrics on several workloads. Am I missing something here @pablogsal ?

pablogsal:
> e a workload called gcbench we can reference. I stand to be corrected but I am not familiar with folks writing a specific workload for every metric they want to measure. Rather its applying some metrics on several workloads. Am I missing something here @pablogsal ?

is very difficult to get a benchmark that is heavily exercising the GC. For now I added these benchmarks to ensure that we can keep track of several aspects of the **CPython** GC that we may want to track, such as how efficient traversal and removing cycles are.

> then things like traversal, collection are just metrics

These things matter for the runtime, so these benchmarks allow us to keep track of them. That's why I am adding them.

> The PyPy folks have a workload called gcbench we can reference

You mean this? https://github.com/mozillazg/pypy/blob/master/rpython/translator/goal/gcbench.py

yes, I plan to add it after some modifications

nanjekyejoannah:
>That's ok, we are not researching GCs, we are keeping track of optimizations in CPython 
> and we want to keep track on how these things get faster for our specific GC, 
> at least for the time being.

I understand the motivation, I will give an example, treating traversal as just a metric, we should
be able to measure it on any workload, typically any  benchmark like pidigits in the pyperformance suite. I am
not very conviced it should be a workload unless you have decided no, only this workload should be 
used to measure GC traversal. Pyperformance workloads process cycles too, albeit not many so a more
rigorous workload is required.

I am not too bent on this, its just appeared wierd to me at first glance, so your decision.

> You mean this? https://github.com/mozillazg/pypy/blob/master/rpython/translator/goal/gcbench.py

Yes, but I maybe biased, its usually sufficient for GC things for us, you can modify if you want.

pablogsal:
> Yes, but I maybe biased, its usually sufficient for GC things for us, you can modify if you want.

When I measured this benchmark had quite a lot of GC activity, but also a lot of other things regarding the VM. The problem is that if the VM gets faster (but not the GC) so it will this benchmark so we won't know where the advantage is coming from. This is why we want to track this separately.

nanjekyejoannah (commented):
I also consider and wish we get the Pyperformance suite to a level of being as agnostic as possible because its considered the stardard Python benchmark suite. If we have unique benchmarks specific to CPython, we can folder them differently as internal.

PR #251, opened 2022-12-03
.github: Add Automated release process

corona10:
* Enable easier release process, we already use the same process for the pyperf releasing.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,28 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --user --upgrade setuptools wheel
+    - name: Build
+      run: |
+        python setup.py sdist bdist_wheel
+    - name: Publish distribution 📦 to PyPI
+      if: startsWith(github.event.ref, 'refs/tags') || github.event_name == 'release'
+      uses: pypa/gh-action-pypi-publish@master
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_API_TOKEN }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,3 +1,5 @@
+# For more information, see the "Release a new version" notes in pyproject.toml.
+
 name: Upload Python Package
 
 on:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,7 +2,7 @@ name: Upload Python Package
 
 on:
   release:
-    types: [created]
+    types: [published]
 
 jobs:
   deploy:
@@ -14,15 +14,17 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: '3.x'
+        cache: pip
+        cache-dependency-path: pyproject.toml
     - name: Install dependencies
       run: |
-        python -m pip install --user --upgrade setuptools wheel
+        python -m pip install --upgrade build
     - name: Build
       run: |
-        python setup.py sdist bdist_wheel
+        python -m build
     - name: Publish distribution 📦 to PyPI
       if: startsWith(github.event.ref, 'refs/tags') || github.event_name == 'release'
-      uses: pypa/gh-action-pypi-publish@master
+      uses: pypa/gh-action-pypi-publish@release/v1
       with:
         user: __token__
         password: ${{ secrets.PYPI_API_TOKEN }}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,12 +16,20 @@
 #  - check the CI status:
 #    https://github.com/python/pyperformance/actions
 #
-# Release a new version:
+# Release a new version with GitHub (preferred):
 #
 #  - go to the GitHub release tab: https://github.com/python/pyperformance/releases
 #  - click "Draft a new release" and fill the contents
-#  - finally click the "Publish release" button! Done!  (see .github/workflows/publish.yml for what gets run)
+#  - finally click the "Publish release" button! Done!
 #  - monitor the publish status: https://github.com/python/pyperformance/actions/workflows/publish.yml
+#
+# Release a new version manually:
+#
+#  - git tag VERSION
+#  - git push --tags
+#  - Remove untracked files/dirs: git clean -fdx
+#  - python -m build
+#  - twine upload dist/*
 #
 # After the release:
 #
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@
 #
 #  - go to the GitHub release tab: https://github.com/python/pyperformance/releases
 #  - click "Draft a new release" and fill the contents
-#  - finally click the "Publish release" button! Done!
+#  - finally click the "Publish release" button! Done!  (see .github/workflows/publish.yml for what gets run)
 #  - monitor the publish status: https://github.com/python/pyperformance/actions/workflows/publish.yml
 #
 # After the release:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,11 +18,10 @@
 #
 # Release a new version:
 #
-#  - git tag VERSION
-#  - git push --tags
-#  - Remove untracked files/dirs: git clean -fdx
-#  - python -m build
-#  - twine upload dist/*
+#  - go to the GitHub release tab: https://github.com/python/pyperformance/releases
+#  - click "Draft a new release" and fill the contents
+#  - finally click the "Publish release" button! Done!
+#  - monitor the publish status: https://github.com/python/pyperformance/actions/workflows/publish.yml
 #
 # After the release:
 #


.github/workflows/publish.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      run: |
+        python -m pip install --user --upgrade setuptools wheel
+    - name: Build
+      run: |
+        python setup.py sdist bdist_wheel
```
hugovk, line 22: Calling setup.py directly is deprecated (long version: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html)

It's recommended to use `build` instead (short version: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html#summary)

```suggestion
    - name: Install dependencies
      run: |
        python -m pip install --upgrade build
    - name: Build
      run: |
        python -m build
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Publish distribution 📦 to PyPI
+      if: startsWith(github.event.ref, 'refs/tags') || github.event_name == 'release'
+      uses: pypa/gh-action-pypi-publish@master
```
hugovk, line 25: Using the `master` branch has been sunset, let's use `release/v1` instead: 

https://github.com/pypa/gh-action-pypi-publish#-master-branch-sunset-

```suggestion
      uses: pypa/gh-action-pypi-publish@ release/v1
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.x'
```
hugovk, line 16: Let's cache pip downloads to help reduce the load on PyPI:

https://dustingram.com/articles/2021/04/14/powering-the-python-package-index-in-2021/

```suggestion
      uses: actions/setup-python@v4
      with:
        python-version: '3.x'
        cache: pip
        cache-dependency-path: setup.py
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,28 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
```
hugovk, line 5: I use `published` here.

Does `created` trigger already when creating a draft release?

If so, is it "safer" to use `published`?

```suggestion
    types: [published]
```

setup.py
```diff
@@ -24,11 +24,10 @@
 #
 # Release a new version:
 #
-#  - git tag VERSION
-#  - git push --tags
-#  - Remove untracked files/dirs: git clean -fdx
-#  - python3 setup.py sdist bdist_wheel
-#  - twine upload dist/*
+#  - go to the GitHub release tab: https://github.com/psf/pyperf/releases
```
hugovk, line 27:
```suggestion
#  - go to the GitHub release tab: https://github.com/python/pyperformance/releases
```

corona10: nah thanks!

hugovk (approved):

corona10:
@ericsnowcurrently PTAL :)

corona10:
@hugovk 
I wrongly rebase the previous commit, so I re-fix your suggetion.
Can you please take a look one more time?

pyproject.toml
```diff
@@ -18,11 +18,10 @@
 #
 # Release a new version:
 #
-#  - git tag VERSION
-#  - git push --tags
-#  - Remove untracked files/dirs: git clean -fdx
-#  - python -m build
-#  - twine upload dist/*
+#  - go to the GitHub release tab: https://github.com/python/pyperformance/releases
+#  - click "Draft a new release" and fill the contents
```
ericsnowcurrently, line 22: Do you mean "Create a new release"?

The URL is https://github.com/python/pyperformance/releases/new, right?

ericsnowcurrently: Regarding "fill the contents", presumably you mean the following:

1. select from "Choose a tag" dropdown (or "create a new tag" via that dropdown)
2. select "Target: main" from the other dropdown (the default?)
3. enter the version number for "Release title", matching the name of the tag from (1)
4. enter something in "Describe this release"

Is the above correct?

Regarding (1), what's the best way to determine what the next version should be?  It might be obvious to many, but it would help some if this were spelled out.  Also, the tag should not exist yet right?

Regarding (4), what should go here?  It seems like, at most, it would include either a basic "what's new" list or a link to a changelog.  Leaving it blank would probably be fine too.  Is there a way to have a template for this field (and maybe for (3) the "Release title" field?

Anyway, what you have is probably fine, but it may be worth putting a bit more info here in this comment in pyproject.toml.

pyproject.toml
```diff
@@ -18,11 +18,10 @@
...
+#  - finally click the "Publish release" button! Done!
+#  - monitor the publish status: https://github.com/python/pyperformance/actions/workflows/publish.yml
```
ericsnowcurrently, line 24: Do you mean that going to that URL will show the status?  Cool!

pyproject.toml
ericsnowcurrently: Also, am I correct in thinking that the "Prepare a release" steps remain unchanged?

If we can have a "publish" template, it would be nice if also reminded the person to make sure they completed those steps already.

pyproject.toml
ericsnowcurrently, line 24: FWIW, it may be worth keeping the old steps as an alternative:

```suggestion
# Release a new version with GitHub (preferred):
#
#  - go to the GitHub release tab: https://github.com/python/pyperformance/releases
#  - click "Draft a new release" and fill the contents
#  - finally click the "Publish release" button! Done!
#  - monitor the publish status: https://github.com/python/pyperformance/actions/workflows/publish.yml
#
# Release a new version manually:
#
#  - git tag VERSION
#  - git push --tags
#  - Remove untracked files/dirs: git clean -fdx
#  - python -m build
#  - twine upload dist/*
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
+name: Upload Python Package
```
ericsnowcurrently, line 1: It may be worth adding a comment pointing to the information in pyproject.toml.

```suggestion
# For more information, see the "Release a new version" notes in pyproject.toml.

name: Upload Python Package
```

pyproject.toml
```diff
@@ -18,11 +18,10 @@
...
 # Release a new version:
 #
-#  - git tag VERSION
-#  - git push --tags
-#  - Remove untracked files/dirs: git clean -fdx
-#  - python -m build
-#  - twine upload dist/*
+#  - go to the GitHub release tab: https://github.com/python/pyperformance/releases
+#  - click "Draft a new release" and fill the contents
+#  - finally click the "Publish release" button! Done!
```
ericsnowcurrently, line 23: Also, it might be nice to point to the workflow file here, so things seem less magical.

```suggestion
#  - finally click the "Publish release" button! Done!  (see .github/workflows/publish.yml for what gets run)
```

ericsnowcurrently (approved):
Thanks for doing this!

LGTM, with some recommendations and a few questions.

I'm not super familiar with the GitHub machinery, so it took me a little while to understand how the workflow maps onto the old steps.  I've got it now, though. :)  The workflow files seems correct, as far as I understood it.

I've left some suggestions about additional/updated comments (and maybe adding a template, if possible), but don't feel like those changes are necessary.  On the one hand, the things I suggested would only really benefit the people with access to publish releases, who are probably already aware of all the information I suggested you explain.  On the other hand, the extra information would probably benefit anyone who receives access to publish but hasn't ever done it before.  (The info would also be useful for folks looking for examples of automated releases through GitHub, though that is more of a positive side effect than a good reason to add all the extra text, etc.)

(So feel free to merge without making any further changes in response to my suggestions, if you think that's okay.)

One other thing that came to mind is if there is any way to automate any of the rest of the release process (e.g. the "Prepare a release" steps)?  What about having GitHub automatically fill in the version number (or even create the tag)?

Regardless of all my thoughts above, what you already have in this PR is a great improvement and a essential foundation for any additional improvements we might consider (though we'll already be better off even if we never make another change related to the release process after this.)  FWIW, I'm certainly not expecting this PR to represent the final desired outcome.. 😄

pyproject.toml
hugovk: For (1), you can either pick an existing tag, or create a new one - the drop down box has "Find or create a new tag". Because we've not created one yet with these steps, we want to create a new one?

For (2) and (3), sounds good.

For (4), one option is to click the "Generate release notes" button and it'll fill something in based on the PRs that were merged.

pyproject.toml
hugovk: Yep! Demo at the other repo: https://github.com/psf/pyperf/actions/workflows/publish.yml

hugovk (approved):
Looks good 👍

hugovk:
> One other thing that came to mind is if there is any way to automate any of the rest of the release process (e.g. the "Prepare a release" steps)? What about having GitHub automatically fill in the version number (or even create the tag)?

Oh yes, there's a lot more automation possible!

Here's roughly what I've set up for my projects (for example https://github.com/hugovk/norwegianblue/).

For PRs, I have to add a "changelog: X" label, where X is one of the types of changes at https://keepachangelog.com/en/1.0.0/: Added, Changed, Deprecated, etc.

For example: https://github.com/hugovk/norwegianblue/pull/116

(I also have automation to ensure such a label has been added to a PR, and automation to add the labels to the repo in the first place.)

Then, I'm using https://github.com/release-drafter/release-drafter to automatically draft a release. It runs when a PR is merged, and lists them by PR title (so I sometimes edit the PR title to make it a good release note title), and also groups them under each "changelog: X" category.

For example: https://github.com/hugovk/norwegianblue/releases/tag/0.9.0

There are often PRs that don't touch production code, or that the end user doesn't care about (e.g. tests or config changes), and I label those with "changelog: skip" so they're left out from the notes.

I have Release Drafter set up to follow SemVer: for bugs we just get a patch bump, for additions/changes we get a minor bump, and for removals we get a major bump.

Also, when PRs are merged to `main`, I deploy to TestPyPI. This keeps the release automation oiled to prevent surprises on release day:

https://github.com/hugovk/norwegianblue/blob/main/.github/workflows/deploy.yml#L46-L51

When I'm ready to release, I go to the release draft at https://github.com/hugovk/norwegianblue/releases. The draft is only shown to those with write access, until published, and currently looks like:

image

I edit the draft, check the version is good; I can change my mind and put another version if I want. I can also adjust what's listed if I want, change order, add/remove extra text etc.

When happy, click "Publish release". This automatically adds the tag, creates the GH release, and kicks off the deploy to PyPI.

I document the release process in https://github.com/hugovk/norwegianblue/blob/main/RELEASING.md so I have a good reference guide (in case another repo has slight variation).

In short: add labels to PRs -> merge PR -> auto drafts release notes -> auto deploy to TestPyPI -> edit release -> publish release -> auto deploy to PyPI.

ericsnowcurrently:
Thanks for working on this, @corona10!  It's a great improvement.

PR #252, opened 2022-12-03
Add the gc benchmarks to the MANIFEST file

pablogsal:


pablogsal:
We should add some check to the CI to ensure these files are not out of sync

PR #254, opened 2022-12-13
Add `asyncio_tcp` benchmark

kumaraditya303:
From https://github.com/python/cpython/issues/91166

We can show off this optimization as part of 3.12 performance improvement :)

### DIFF
--- a/pyperformance/data-files/benchmarks/bm_asyncio_tcp/run_benchmark.py
+++ b/pyperformance/data-files/benchmarks/bm_asyncio_tcp/run_benchmark.py
@@ -1,0 +1,41 @@
+"""
+Benchmark for asyncio TCP server and client performance
+transferring 10MB of data.
+
+Author: Kumar Aditya
+"""
+
+
+import asyncio
+from pyperf import Runner
+
+
+CHUNK_SIZE = 1024 ** 2 * 10
+
+
+async def handle_echo(reader: asyncio.StreamReader,
+                      writer: asyncio.StreamWriter) -> None:
+    data = b'x' * CHUNK_SIZE
+    for _ in range(100):
+        writer.write(data)
+        await writer.drain()
+    writer.close()
+    await writer.wait_closed()
+
+
+async def main() -> None:
+    server = await asyncio.start_server(handle_echo, '127.0.0.1', 8882)
+
+    async with server:
+        asyncio.create_task(server.start_serving())
+        reader, writer = await asyncio.open_connection('127.0.0.1', 8882)
+        while True:
+            data = await reader.read(CHUNK_SIZE)
+            if not data:
+                break
+        writer.close()
+        await writer.wait_closed()
+
+if __name__ == '__main__':
+    runner = Runner()
+    runner.bench_async_func('asyncio_tcp', main)
--- a/pyperformance/data-files/benchmarks/bm_asyncio_tcp/run_benchmark.py
+++ b/pyperformance/data-files/benchmarks/bm_asyncio_tcp/run_benchmark.py
@@ -29,13 +29,13 @@ async def main() -> None:
     async with server:
         asyncio.create_task(server.start_serving())
         reader, writer = await asyncio.open_connection('127.0.0.1', 8882)
-        received = 0
+        data_len = 0
         while True:
             data = await reader.read(CHUNK_SIZE)
             if not data:
                 break
-            received += len(data)
-        assert received == CHUNK_SIZE * 100
+            data_len += len(data)
+        assert data_len == CHUNK_SIZE * 100
         writer.close()
         await writer.wait_closed()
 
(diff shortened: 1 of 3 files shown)


brandtbucher:
Is the goal for this benchmark to guide our future improvements, or just to highlight the improvement already made in the linked issue?

We should really avoid the latter... the last thing I want is to add synthetic benchmarks just to boost our 3.12 numbers.

kumaraditya303:
> Is the goal for this benchmark to guide our future improvements, or just to highlight the improvement already made in the linked issue?

It is to guide for future improvements and more importantly avoid regressions.  As it currently stands there are no benchmarks for `asyncio` thus it is hard to evaluate the impact of any bug fixes on performance, having this is `pyperformance` will help in tracking this.  (recent https://github.com/python/cpython/pull/100053)

> We should really avoid the latter... the last thing I want is to add synthetic benchmarks just to boost our 3.12 numbers.

That's true though this isn't any different from other benchmark, even if there is improvement in `pyperformance` it doesn't mean the user's application will be that much faster or faster at all. 

I am fine with keeping this in my own fork but I would prefer to have it in `pyperformance` so it can be tracked using speed.python.org.

kumaraditya303:
FYI if this is acceptable then I have other benchmarks to add for `asyncio` notably `udp` and `ssl` performance.

corona10:
Comment which does not judge this PR: I think that adding a new benchmark will consume the cost of running time for measuring.
Since we can not add every single benchmark for each dimension, each benchmark should cover a lot of dimensions of performance if possible.

kumaraditya303:
> I think that adding a new benchmark will consume the cost of running time for measuring.

I was under the impression that cost of running the benchmark is not an issue. If it is an issue then it isn't worth it and I can run it myself when necessary so no need for this in pyperformance.

ericsnowcurrently:
> I think that adding a new benchmark will consume the cost of running time for measuring.
> Since we can not add every single benchmark for each dimension, each benchmark should cover a lot of dimensions of performance if possible.

If running time is a concern, we can restrict the set of benchmarks that are run by default.  That is easily done in the [manifest file](https://github.com/python/pyperformance/blob/main/pyperformance/data-files/benchmarks/MANIFEST).  Then we can do longer runs whenever we feel comfortable with the extra time by passing the `--benchmarks all` (or maybe it is "<all>"?) option to pyperformance.

In the long term, ideally we would have a solid set of workload-oriented ("macro") benchmarks that measure real-world performance (close enough, at least).  This is what we would use to measure performance effectively and those results are what we would communicate to users.  However, we still need the wide variety of feature-oriented ("micro") benchmarks, to help us avoid regressions and narrow down causes when we have regressions, as well as point to where possible improvements might be.  See https://github.com/faster-cpython/ideas/wiki/All-About-Python-Benchmarking.

Currently, pretty much all of the pyperformance benchmarks are feature-oriented ("micro").  So we have plenty of work ahead of us.  (Note that the [pyston benchmarks](https://github.com/pyston/python-macrobenchmarks/tree/main/benchmarks) are much closer to workload-oriented and we have been incorporating them into the faster-cpython results as much as possible.)

corona10:
> If running time is a concern, we can restrict the set of benchmarks that are run by default.

Ah okay, in that case, I am okay to add it.
I just worried about all various of bench_asyncio_xxxx.py.

kumaraditya303:
Final benchmark result: 

| Benchmark   | 3.10 | patch                |
|-------------|:--------:|:--------------------:|
| asyncio_tcp | 1.41 sec | 724 ms: 1.95x faster |

I plan to merge this by weekend if there are no objections.

pyperformance/data-files/benchmarks/bm_asyncio_tcp/run_benchmark.py
```diff
@@ -0,0 +1,41 @@
...
+async def main() -> None:
+    server = await asyncio.start_server(handle_echo, '127.0.0.1', 8882)
+
+    async with server:
+        asyncio.create_task(server.start_serving())
+        reader, writer = await asyncio.open_connection('127.0.0.1', 8882)
+        while True:
+            data = await reader.read(CHUNK_SIZE)
+            if not data:
+                break
```
gvanrossum, line 35: I feel this should at least count the total number of bytes received and assert that it is as expected (100*CHUNK_SIZE?).

kumaraditya303: done

kumaraditya303: For some reason I didn't receive notification for your comment :/.

gvanrossum (approved):

gvanrossum:
@mdboom What do you think?

kumaraditya303:
I'll merge this now as https://github.com/python/cpython/pull/31871 is almost done, awaiting final review and merge. We can sort out any issues with this later down the road.

PR #258, opened 2023-01-10
Fix parameter

penguin-wwy:
An undefined variable is used as a parameter, which looks like it should be wrong.


StanFromIreland:
Sorry we missed this, it has been fixed in https://github.com/python/pyperformance/commit/f3c44951a533dc4835f8ba9d8b6f2ea069ad79ed.

PR #259, opened 2023-01-11
Skip 3.12-dev pending Greenlet support

hugovk:
At the moment the `3.12-dev` builds are failing, causing every build to show as failed.

It's failing because Greenlet doesn't yet support 3.12. The tracking issue is at https://github.com/python-greenlet/greenlet/issues/323 and there's a PR at https://github.com/python-greenlet/greenlet/pull/327 but it's not ready yet.

In the meantime, we can mark `3.12-dev` as "experimental", so whilst the individual job still fails, it doesn't mark the whole build as failed.

Docs: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

---

Also bump GitHub Actions, allow building feature branches, and we can fold the PyPy config into the main list. `workflow_dispatch` adds a button to the UI to be able to trigger new builds, occasionally useful.

---

Shall we also remove `pypy3.7`? The newest PyPy only supports 3.8 and 3.9:

* https://www.pypy.org/posts/2022/12/pypy-v7311-release.html

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,7 +19,7 @@ jobs:
         # Test all supported versions on Ubuntu:
         os: [ubuntu-latest]
         # Skip 3.12-dev, pending https://github.com/python-greenlet/greenlet/issues/323
-        python: [pypy3.7, pypy3.8, pypy3.9, "3.7", "3.8", "3.9", "3.10", 3.11-dev]
+        python: [pypy3.8, pypy3.9, "3.7", "3.8", "3.9", "3.10", 3.11-dev]
         include:
           # Also test macOS and Windows:
           - os: macos-latest


hugovk:
I'm wondering if this needs to merged first? 

At least it shows as green at https://github.com/python/pyperformance/actions

image

hugovk:
Need to research more about how `continue-on-error` works, for now I think it's fine to remove the failing `3.12-dev` from the matrix until https://github.com/python-greenlet/greenlet/issues/323 is resolved.

I also removed unsupported `pypy3.7`.

hugovk:
Please see PR https://github.com/python/pyperformance/pull/264 to replace this.

.github/workflows/main.yml
```diff
@@ -19,28 +18,23 @@ jobs:
       matrix:
         # Test all supported versions on Ubuntu:
         os: [ubuntu-latest]
-        python: ["3.7", "3.8", "3.9", "3.10", 3.11-dev, 3.12-dev]
+        # Skip 3.12-dev, pending https://github.com/python-greenlet/greenlet/issues/323
+        python: [pypy3.8, pypy3.9, "3.7", "3.8", "3.9", "3.10", 3.11-dev]
```
pradyunsg, line 22: If I'm understanding https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures correctly, we can achieve what we want to with something like:

```yaml
    continue-on-error: ${{ matrix.failure-allowed }}
    strategy:
      fail-fast: false
      matrix:
        # Test all supported versions on Ubuntu:
        os: [ubuntu-latest]
        python: [pypy3.8, pypy3.9, "3.7", "3.8", "3.9", "3.10", 3.11-dev]
        failure-allowed: [false]
        include:
          # Also test macOS and Windows:
          - os: macos-latest
            python: "3.10"
            failure-allowed: false
          - os: windows-latest
            python: "3.10"
            failure-allowed: false
          # Also test in-development version, allowing failures
          - os: ubuntu-latest
            python: "3.12-dev"
            failure-allowed: true
```

pradyunsg: `failure-allowed` is a verb/phrase that we can change as we wish to communicate things via the variable name. The example from GitHub used `experimental`.

hugovk: Thanks, I did try something along those lines, but wasn't successful. 

Although I think https://github.com/python/pyperformance/pull/264 is a better approach? Then we'll at least see if the non-greenlet tests fail for 3.12. And when greenlet installs for 3.12, it'll start testing those too.

pradyunsg: Looking at your force-pushed events' diffs, I don't see `continue-on-error: ${{ matrix.failure-allowed }}` being added at any point at the `job.<id>` level; which might be why they didn't work?

Regardless, they're both valid approaches to solving the problem so it's not a big deal which one we go with. :)

hugovk: I also tried it in [another branch](https://github.com/hugovk/pyperformance/blob/ci-update2/.github/workflows/main.yml):

```yml
  test:
    runs-on: ${{ matrix.os }}
    name: ${{ matrix.os }} - ${{ matrix.python }}
    continue-on-error: ${{ matrix.experimental }}
    strategy:
      fail-fast: false
      matrix:
        # Test all supported versions on Ubuntu:
        os: [ubuntu-latest]
        python: [pypy3.7, pypy3.8, pypy3.9, "3.7", "3.8", "3.9", "3.10", 3.11-dev]
        experimental: [false]
        include:
          # Also test macOS and Windows:
          - os: macos-latest
            python: "3.10"
            experimental: false
          - os: windows-latest
            python: "3.10"
            experimental: false
          # Allow 3.12-dev to fail pending https://github.com/python-greenlet/greenlet/issues/323
          - os: ubuntu-latest
            python: 3.12-dev
            experimental: true
```

Very similar to your snippet:

image

But still failed a PR: https://github.com/hugovk/pyperformance/pull/3

pradyunsg: Indeed! Thanks for clarifying! ^.^

hugovk:
Closed in favour of https://github.com/python/pyperformance/pull/274 👍

PR #264, opened 2023-02-05
Skip tests failing on Python 3.12 pending greenlet support

### LINKED ISSUE #263: Tests fail with Python 3.12
Whenever I push something to a PR in the cpython main branch (3.12) I get an email from this repo telling me that the 3.12 run failed. Is there a way we can fix it? Here's a link to the failing run: https://github.com/python/pyperformance/actions/runs/4094033809/jobs/7059879010

I'd rather not just disable running with 3.12 (that would presumably take away some useful signal) but maybe we can fix or suppress the failing tests?

hugovk:
Fixes #263.

Alternative to and closes https://github.com/python/pyperformance/pull/259.

At the moment the `3.12-dev` builds are failing, causing every build to show as failed.

It's failing because greenlet doesn't yet support 3.12. The tracking issue is at https://github.com/python-greenlet/greenlet/issues/323 and there's a PR at https://github.com/python-greenlet/greenlet/pull/327 but it's not ready yet.

In the meantime, let's skip the two tests when greenlet isn't available.

---

Also bump GitHub Actions, allow building feature branches, and we can fold the PyPy config into the main list. `workflow_dispatch` adds a button to the UI to be able to trigger new builds, occasionally useful.

And remove `pypy3.7`, the newest PyPy only supports 3.8 and 3.9:

* https://www.pypy.org/posts/2022/12/pypy-v7311-release.html

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,12 +2,11 @@ name: test
 
 on:
   push:
-    branches: [main]
   pull_request:
-    branches: [main]
   schedule:
     # Midnight UTC:
     - cron: "0 0 * * *"
+  workflow_dispatch:
 
 jobs:
 
@@ -33,14 +32,14 @@ jobs:
           - os: windows-latest
             python: "3.10"
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python }}
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       if: "!endsWith(matrix.python, '-dev')"
       with:
         python-version: ${{ matrix.python }}
     - name: Set up Python ${{ matrix.python }} using deadsnakes
-      uses: deadsnakes/action@v2.1.1
+      uses: deadsnakes/action@v3.0.0
       if: "endsWith(matrix.python, '-dev')"
       with:
         python-version: ${{ matrix.python }}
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -2,12 +2,11 @@ name: test
 
 on:
   push:
-    branches: [main]
   pull_request:
-    branches: [main]
   schedule:
```
pradyunsg, line 6: Let's keep `branches: [main]` from `push`? Also, I was a bit surprised that this was changed in a commit bumping the actions' versions given the granularity of other commits. :)

With this change, any branch pushed to this project will trigger a CI run; which could result in duplicated CI runs for something like a README typo fix submitted by a maintainer.

hugovk: Unfortunately with `push: branches: [main]`, it's not possible to test on our forks before opening PRs, and I wouldn't have contributed this or #259 in the first place.

Looking at recent PRs, all maintainers create PRs from their own forks. For the occasional typo fix made directly in this repo, I think the duplicate build is fine, and it's much more important to test forks.

If it's not fine, how does this config look?

```yml
    # We want to run on external PRs, but not on our own internal PRs as they'll be run
    # by the push to the branch.
    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
```

re: https://github.com/has2k1/plotnine/blob/f696c01f1602e422a29778c112005463992ed707/.github/workflows/testing.yml#L10-L12

pradyunsg: Ah, good point. I guess leaving this as-is works fine; since the cost is a duplicated run which will fail the same was as other runs.

PR #265, opened 2023-02-15
add comprehensions benchmark

carljm:
This benchmark is loosely based on real code in an internal workload where it makes heaviest use of list comprehensions, obviously simplified and anonymized.

### DIFF
--- a/pyperformance/data-files/benchmarks/MANIFEST
+++ b/pyperformance/data-files/benchmarks/MANIFEST
@@ -17,6 +17,7 @@ gc_collect	<local>
 generators	<local>
 chameleon	<local>
 chaos	<local>
+comprehensions	<local>
 crypto_pyaes	<local>
 dask	<local>
 deepcopy	<local>
--- a/pyperformance/data-files/benchmarks/bm_comprehensions/pyproject.toml
+++ b/pyperformance/data-files/benchmarks/bm_comprehensions/pyproject.toml
@@ -1,0 +1,9 @@
+[project]
+name = "pyperformance_bm_comprehensions"
+requires-python = ">=3.10"
+dependencies = ["pyperf"]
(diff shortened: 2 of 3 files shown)


carljm:
The 3.12 failure looks like a greenlet build issue and not related to this PR.

The pre-3.10 failures are expected I guess; I thought putting `requires-python = ">=3.10"` in the `pyproject.toml` for the new benchmark would cause it to be skipped on earlier python versions, but apparently not. I guess I will just remove the 3.10-only type annotations from the benchmark (they aren't really relevant or needed anyway) so it can work on 3.7+.

mdboom (approved):

carljm:
@mdboom is there a particular reviewer you are waiting on before merging this? I requested review from you and @markshannon; happy to have Mark look at it if he has time, but if not, do you feel comfortable going ahead and merging it?

arhadthedev:
@mdboom is a triager, not a core developer so cannot merge.

@kumaraditya303 could you have a look please?

carljm:
Ah, makes sense, thanks. For some reason I thought pyperformance had a different maintenance team from core CPython and that @mdboom was on it :)

kumaraditya303:
LGTM but before merging I would like to see a comparison with one of your PRs which implements comprehensions inlining against main.

kumaraditya303:
> do you have reason to think otherwise? 

I don't disagree, I was just curious of how much speedup inlining can make, anyways I'll merge this now.

carljm:
@kumaraditya303 Thanks for merging! Preliminary numbers last week showed an 11-12% improvement in this benchmark from full inlining (https://github.com/python/cpython/pull/101441). I haven't measured it yet against https://github.com/python/cpython/pull/101310. Now that it's merged I'll do a careful measurement against both of those PRs.

PR #271, opened 2023-04-14
add version of Richards that uses super()

carljm:
Currently, none of the benchmarks in pyperformance exercise `super()` method calls significantly. But real-world modern object-oriented Python code often does (one source: Instagram server codebase uses lots of `super()` method calls, some in hot paths.)

To correct this gap in visibility, this PR adds a copy of the Richards benchmark that is modified to use `super()` calls. Existing super-method calls that were done by naming the base class directly (`Task.__init__(self, ...)`) are modified to use `super()`, and the already-polymorphic `Task.fn` method is modified slightly such that overrides now call up to the base method using `super()`.

If we accept that Richards is a roughly reasonable representation of some object-oriented Python code, I think this is a similarly reasonable representation of the real-world use of `super()` in such code.

Thanks to @pablogsal for this suggestion.

### DIFF
--- a/pyperformance/data-files/benchmarks/bm_richards_super/run_benchmark.py
+++ b/pyperformance/data-files/benchmarks/bm_richards_super/run_benchmark.py
@@ -32,7 +32,7 @@ BUFSIZE = 4
 BUFSIZE_RANGE = range(BUFSIZE)
 
 
-class Packet(object):
+class Packet:
 
     def __init__(self, l, i, k):
         self.link = l
@@ -57,7 +57,7 @@ class Packet(object):
 # Task Records
 
 
-class TaskRec(object):
+class TaskRec:
     pass
 
 
@@ -97,7 +97,7 @@ class WorkerTaskRec(TaskRec):
 # Task
 
 
-class TaskState(object):
+class TaskState:
 
     def __init__(self):
         self.packet_pending = True
@@ -160,7 +160,7 @@ def trace(a):
 TASKTABSIZE = 10
 
 
-class TaskWorkArea(object):
+class TaskWorkArea:
 
     def __init__(self):
         self.taskTab = [None] * TASKTABSIZE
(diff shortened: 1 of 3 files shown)


carljm:
The 3.12 failure looks like an issue building greenlet, not related to this PR.

hugovk:
Yep, see https://github.com/python/pyperformance/issues/263 for 3.12+greenlet.

corona10:
Nice I will take a look

corona10:
Overall looks good, but why not just replace the Richards to use super()?
I feel the running time of benchmarks is increased due to the increasing number of benchmarks.
Adding benchmarks is good, but let's avoid the duplicated similar workload added if possible.

cc @pablogsal

carljm:
> Overall looks good, but why not just replace the Richards to use super()?

I think the main reason is because this means Richards timing is no longer comparable to previous timings?

pyperformance/data-files/benchmarks/bm_richards_super/run_benchmark.py
```diff
@@ -0,0 +1,430 @@
...
+K_WORK = 1001
+
+# Packet
+
+BUFSIZE = 4
+
+BUFSIZE_RANGE = range(BUFSIZE)
+
+
+class Packet(object):
```
AlexWaygood, line 35: Since this is a new benchmark, and we won't be comparing the results of this benchmark to the results of the old benchmark, I would remove explicit inheritance from `object` in benchmark, so that it looks like idiomatic, modern Python code:

```suggestion
class Packet:
```

(etc. throughout the PR)

mdboom:
This looks good to me.  I agree that adding (rather than replacing) is the way to go here.  We can always remove the old benchmark once we learn more from this one.

mdboom (approved):
LGTM.

corona10 (approved):
lgtm

PR #272, opened 2023-04-14
Upgrade to pyperf 2.6.0

itamaro:
needed for loop_factory parameter support in bench_async_func


hugovk:
The 3.12 failure is a known issue (https://github.com/python/pyperformance/issues/263) but please could you check the new PyPy failures?

itamaro:
thanks @hugovk, I'm actually not sure what these failures are about, nothing jumps at me from the logs.
could you trigger retries perhaps for the pypy jobs?

hugovk:
Restarted, and they fail in the same way.

itamaro:
Hmm that's odd. I'll need to look closer and get a local repro.

corona10:
> Hmm that's odd. I'll need to look closer and get a local repro.

pyperf doesn't have CI for PyPy yet. So I will add CIs for PyPy to pyperf first.

corona10:
I ran the pyperformance benchmark, and I got a different exit code per pyperf version

* command: $> pypy -m pyperformance run -b all --debug-single-value -o foo.json
* pyperf 2.6.0: echo $? -> 0
* pyperf 2.7.0: echo $? -> 1

There might be some regression from pyperf 2.7.0

corona10:
@hugovk 
Okay, pypy failed for running several tests, and the change of pyperf makes a different output (I verified the issue by reverting the specific patch)
: https://github.com/psf/pyperf/commit/d3f3e6a4ce2a59fe26408e434561182fa8bc3474

Two ways to solve
1. Revert pyperf change and release pyperf 2,6,1
2. Set pypy CI as the experimental task

Which do you prefer? I think that the pyperf 2.6.0 is doing what it should do.

itamaro:
If 2.6.0 is behaving correctly, could we update the test to match the output?

corona10:
> If 2.6.0 is behaving correctly, could we update the test to match the output?

From that side of view, since pypy fails to run the pyperformance benchmark suite,
There are two ways to solve.
1. If the benchmark is wrong benchmark code should be updated.
2. If the pypy implementation issue, then pypy should update their implementation (we already did for CPython 3.11 and 3.12)

And I prefer to update GHA file instead of touching pyperformance code directly if not, we have to release the pyperformance with considering the implementation issue.
https://github.com/python/pyperformance/blob/c4fcdd98141079988280180fd1a814d48866824f/.github/workflows/main.yml#L34-L41

Then we don't have to consider the implementation issue when we release the pyperformance.

corona10:
cc @cfbolz

cfbolz:
ok, found it. it's not really a pypy bug as such (ie it would affect all the python implementations that are counted as having a jit) but a weird interaction of things.

 the direct cause is this `pyperf` change: https://github.com/psf/pyperf/commit/d5349bb409cc94eaca6a6646ca5df9010e3f94db#diff-d1600cb0c5deea10c84136091897ee5086097ac9e35be850235c1f28b7a48316L74

it removes the `warmups` argument to `Runner.__init__`. This parameter has not actually *done* anything in a [long time](https://github.com/psf/pyperf/commit/d21fee833b26a6170a2e5e2533821d2e9a650b0b#diff-80086baf88b49811bed7ea6f51429845ec5c6494f2c5ed0214f94add87631e54L140). However, it is still passed into the constructor in three benchmarks:

- [go](https://github.com/python/pyperformance/blob/main/pyperformance/data-files/benchmarks/bm_go/run_benchmark.py#L458)
- [hexiom](https://github.com/python/pyperformance/blob/main/pyperformance/data-files/benchmarks/bm_hexiom/run_benchmark.py#L657)
- [tornado_http](https://github.com/python/pyperformance/blob/main/pyperformance/data-files/benchmarks/bm_tornado_http/run_benchmark.py#L102)

my suggestion is to simply remove the `warmups` use in these three benchmarks.

corona10:
> my suggestion is to simply remove the warmups use in these three benchmarks.

@cfbolz 
Thanks for your investigation!!
In that case, it will be 'If the benchmark is wrong, benchmark code should be updated.' case. :)


@itamaro 
Would you like to follow what @cfbolz suggested and let's see if it changes to green status.

itamaro:
Thanks @cfbolz , all checks are green now!

itamaro:
Thanks @cfbolz , all checks are green now!

corona10 (approved):
lgtm

PR #277, opened 2023-04-21
.github: Set pypy CI as the experimental mode

corona10:
ref: https://github.com/python/pyperformance/pull/272#issuecomment-1517845909


corona10:
cc @cfbolz

cfbolz:
it's not pypy's fault, so I am -1 here. see https://github.com/python/pyperformance/pull/272#issuecomment-1518622125

PR #279, opened 2023-04-21
Extend async tree benchmarks to cover eager task execution

itamaro:
goes along with https://github.com/python/cpython/issues/97696 (implemented in https://github.com/python/cpython/pull/102853)

this change to pyperformance was used to measure the impact of the eager task execution feature.

### DIFF
--- a/pyperformance/data-files/benchmarks/bm_async_tree/bm_async_tree_cpu_io_mixed.toml
+++ b/pyperformance/data-files/benchmarks/bm_async_tree/bm_async_tree_cpu_io_mixed.toml
@@ -1,4 +1,3 @@
 [tool.pyperformance]
 name = "async_tree_cpu_io_mixed"
 extra_opts = ["cpu_io_mixed"]
-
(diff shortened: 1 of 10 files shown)


willingc:
LGTM. Thanks @itamaro

willingc (approved):

itamaro:
with https://github.com/python/cpython/pull/102853 merged, I think this PR can be merged now so we can continue benchmarking this feature

pyperformance/data-files/benchmarks/bm_async_tree/bm_async_tree_cpu_io_mixed.toml
```diff
@@ -1,4 +1,3 @@
 [tool.pyperformance]
 name = "async_tree_cpu_io_mixed"
 extra_opts = ["cpu_io_mixed"]
-
```
kumaraditya303, line 4: Please keep one trailing line.

itamaro: main has two trailing lines here (and in the other one) - GitHub UI is not super clear about this, but the change here is from two trailing lines to one trailing line (unless my editor is messing with me).

pyperformance/data-files/benchmarks/bm_async_tree/run_benchmark.py
```diff
@@ -57,16 +60,32 @@ async def run(self):
         await self.recurse(NUM_RECURSE_LEVELS)
 
 
+class EagerMixin:
+    async def run(self):
+        loop = asyncio.get_event_loop()
```
kumaraditya303, line 65:
```suggestion
        loop = asyncio.get_running_loop()
```

Issue #105, opened 2021-07-31
Improving representative benchmarks for typing ecosystem


Fidget-Spinner:
Due to a current lack of representative macrobenchmarks, it is very difficult to give judgement for more complex accelerators for some parts of typing. Hence, I'm trying to upstream some benchmarks into pyperformance.

IMO, there are three main areas:
1. Performance of static type checkers implemented in Python (e.g. mypy). (Fixed by #102)
2. Performance of programs using types at runtime (e.g. pydantic, attrs, etc.).
3. Runtime overhead of typed code vs fully untyped code.

For case 2, I plan to use one of pydantic's benchmarks here https://github.com/samuelcolvin/pydantic/tree/master/benchmarks.

Case 3 is very tricky because there are so many ways to use typing. I don't know how often people use certain features, whether they type-hint inside tight loops, etc. So I'm struggling to find a good benchmark. An idea: grabbing one of the existing pyperformance benchmarks, fully type-hinting it, then comparing the performance delta may work.

CC @JelleZijlstra, I would greatly appreciate hearing your opinion on this (especially for case 3). Maybe I can post this on typing-sig too if I need more help.

Afteword:
All 3 cases benefit from general CPython optimizations. But usually only 3. benefits greatly from typing-only optimizations (with 1. maybe not improving at all).

JelleZijlstra:
Here are some thoughts:
* mypy itself is actually also a good example of a fully typed Python codebase that's friendly to benchmarking. Comparing its performance with all types stripped out (but still in pure Python mode, no mypyc) could be interesting.
* I think there's a lot of variation in how people use typing. For example, my company's codebase very heavily uses NewTypes as annotations, but barely uses generics. But mypy doesn't use NewTypes internally but uses a lot of generics.
* Here are some aspects of static typing that could plausibly affect runtime performance
   * Import time cost (in both speed and memory) of evaluating lots of annotations. This was a major part of the motivation for PEP 563 and PEP 649. The latter is still under consideration by the SC, so benchmarks could help inform a decision.
   * Instantiation of generic classes. I remember this was especially an issue early on but I think Ivan (?) made some fixes later in the 3.x series. (So `class X(Generic[T]): ...` made `X()` slow.) This would be a good benchmark to add.
[...]

PR #280, opened 2023-04-25
Add a benchmark for runtime-checkable protocols

AlexWaygood:
We had a longstanding issue open at CPython complaining that the performance of `isinstance()` checks against runtime-checkable protocols was a serious bottleneck for their programme. The performance of these `isinstance()` checks has been significantly improved for 3.12 compared to 3.11, but there's currently no code (to my knowledge) in pyperformance that significantly exercises `isinstance()` checks against runtime-checkable protocols.

Work towards #105.

### DIFF
--- a/pyperformance/data-files/benchmarks/MANIFEST
+++ b/pyperformance/data-files/benchmarks/MANIFEST
@@ -69,6 +69,7 @@ sympy	<local>
 telco	<local>
 tomli_loads	<local>
 tornado_http	<local>
+typing_runtime_protocols	<local>
 unpack_sequence	<local>
 unpickle	<local:pickle>
 unpickle_list	<local:pickle>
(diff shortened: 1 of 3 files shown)


hauntsaninja (commented):
Might be worth having more than five. I just grepped and I found a runtime_checkable Protocol in a codebase at work with twenty one methods.....

AlexWaygood:
> Might be worth having more than five. I just grepped and I found a runtime_checkable Protocol in a codebase at work with twenty one methods.....

I did a fair amount of grepping using grep.app while writing this benchmark. The vast majority of runtime-checkable protocols in the wild appear to have 5 members or fewer. However, there are a few codebases that make _heavy_ use of runtime-checkable protocols, and they often have "monster protocols" like the one you describe, which may have truly awful performance (and will probably have worse performance in 3.12 than they did in 3.11).

So, I'm not sure what the best thing to do here is!

brandtbucher (commented):
Thanks for taking the time to do this, and for the helpful comment! I have a couple of questions/suggestions:

AlexWaygood:
> Thanks for taking the time to do this, and for the helpful comment! I have a couple of questions/suggestions:

Thanks for the review! I think I addressed all the comments.

brandtbucher (approved):

PR #283, opened 2023-04-26
On older versions of Python, skip benchmarks that use features introduced in newer Python versions

### LINKED ISSUE #281: Impossible to add a benchmark that exercises newer Python features
It appears to be currently impossible to add a benchmark that uses a feature added in Python >3.7. Even if you add the `requires-python` field in the `pyproject.toml` file for a specific benchmark, the benchmark is run on all Python versions in CI. This means that the CI will fail if the benchmark includes a feature that is new in 3.8, for example.

This issue currently blocks both of these open PRs:
- #268
- #280

AlexWaygood:
All our benchmarks have a `requires-python` field in their `pyproject.toml` files, e.g.: https://github.com/python/pyperformance/blob/974e29c54048e3a980b20a49a3ddd114e1b93892/pyperformance/data-files/benchmarks/bm_2to3/pyproject.toml#L3

The `requires-python` field is added to the metadata of each benchmark as a `python` field here:

https://github.com/python/pyperformance/blob/974e29c54048e3a980b20a49a3ddd114e1b93892/pyperformance/_benchmark_metadata.py#L22

https://github.com/python/pyperformance/blob/974e29c54048e3a980b20a49a3ddd114e1b93892/pyperformance/_benchmark_metadata.py#L193-L202

We can use that metadata to create a new `python` property on `Benchmark` objects, which returns a `packaging.specifiers.SpecifierSet` instance. This property can then be easily used to filter out benchmarks that require a higher version of Python than the version of Python pyperformance is running on.

Fixes #281. Unblocks #280 and #268.

I haven't added a test for this -- I was unsure if it was necessary, and if it was necessary, where that test should go. I'm happy to add one if that would be helpful and there's an obvious place it could go, however!

I manually tested adding a benchmark that used new-in-Python-3.8 features in 1734919c7ca476aba40b03121e310467878a2702, and the CI passed fine (adding the same benchmark to `main` currently causes the CI to fail — see #280). Passing CI run on my GitHub fork: https://github.com/AlexWaygood/pyperformance/actions/runs/4813903519


pyperformance/_benchmark.py
```diff
@@ -164,9 +165,13 @@ def runscript(self):
     def extra_opts(self):
         return self._get_metadata_value('extra_opts', ())
 
+    @property
+    def python(self):
+        req = self._get_metadata_value("python", None)
+        return None if req is None else SpecifierSet(req)
```
brandtbucher, line 171:
```suggestion
        return SpecifierSet(self._get_metadata_value("python", ""))
```

pyperformance/cli.py
```diff
@@ -241,11 +241,15 @@ def parse_entry(o, s):
...
     # Get the selections.
     selected = []
+    this_python_version = ".".join(map(str, sys.version_info[:3]))
     for bench in _benchmark_selections.iter_selections(manifest, parsed_infos):
         if isinstance(bench, str):
             logging.warning(f"no benchmark named {bench!r}")
             continue
-        selected.append(bench)
+        # Filter out any benchmarks that can't be run on the Python version we're running
+        if bench.python is not None and this_python_version in bench.python:
```
brandtbucher, line 250:
```suggestion
        if this_python_version in bench.python:
```

brandtbucher (approved):
I'm impressed how simple this is! Great job.

One opportunity to remove some branching:

pyperformance/_benchmark.py
AlexWaygood: Oh, nice!

Issue #282, opened 2023-04-26
Add type hints?


AlexWaygood:
How would we feel about adding type hints to the pyperformance codebase?

Running a type checker on pyperformance might catch some bugs. But more importantly, in my opinion, it would make some parts of the code base a lot easier to read and reason about -- it would be self-evident what kind of object is being accepted and returned by various functions.

There are various tools that we could use to add "obvious" type hints quickly to the code base, such as [autotyping](https://github.com/JelleZijlstra/autotyping) and [MonkeyType](https://github.com/Instagram/MonkeyType).

PR #284, opened 2023-04-30
Use `tomllib/tomli` over `toml`

AlexWaygood:
`tomli` is more actively maintained than `toml`, is more up to date with the TOML spec, and has been incorporated into the Python stdlib as `tomllib` in Python 3.11.

I think this is worth doing anyway, but this also helps with #282, as `tomli` is a `py.typed` package that ships with inline type annotations.


pyperformance/data-files/requirements.txt
```diff
@@ -2,17 +2,13 @@
...
+psutil==5.9.5
     # via
     #   -r requirements.in
     #   pyperf
-pyparsing==3.0.8
-    # via packaging
 pyperf==2.6.0
     # via -r requirements.in
-toml==0.10.2
-    # via -r requirements.in
```
AlexWaygood, line 18: I ran `pip-compile` to regenerate this file, and it also bumped `packaging` and `psutil`. Not sure if that's desirable or if I did something wrong there!

pyperformance/_pyproject_toml.py
```diff
@@ -12,13 +12,18 @@
...
 import packaging.requirements
 import packaging.specifiers
 import packaging.utils
 import packaging.version
-import toml
+
+if sys.version_info >= (3, 11):
+    import tomllib
+else:
+    import tomli as tomllib
```
corona10, line 26: I prefer to use the ImportError strategy for this code.
We should consider alternative implementations and we can not guarantee the tomlib from those implementations.

AlexWaygood: Thanks, done!

corona10 (approved):
lgtm!

PR #285, opened 2023-05-01
Don't run the daily test on contributors' GitHub forks

AlexWaygood:
I'm currently getting a daily email informing me that the pyperformance tests failed on the `AlexWaygood/pyperformance` repo (because I enabled workflows for my fork). This PR makes it so that the `test` workflow only happens on `python/pyperformance`, not on contributors' forks of the repo (unless they specifically request the workflow by manually triggering the workflow or by making a PR to their fork).

We use a condition like this in our typeshed workflow here, and it works great: https://github.com/python/typeshed/blob/ae0c9f9dadd72ee823d289416943a9109e24fe52/.github/workflows/daily.yml#L33

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,6 +17,7 @@ jobs:
     permissions:
       pull-requests: write
     name: ${{ matrix.os }} - ${{ matrix.python }}
+    if: ${{ github.repository == 'python/pyperformance' || github.event_name != 'schedule' || github.event_name != 'pull_request' }}
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
     permissions:
       pull-requests: write
     name: ${{ matrix.os }} - ${{ matrix.python }}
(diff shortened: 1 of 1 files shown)


hugovk:
Good idea!

```yml
if: ${{ github.repository == 'python/pyperformance' || github.event_name != 'schedule' || github.event_name != 'pull_request' }}
```

So does that mean run:

* on anything for upstream
  OR
* on non-`schedule` triggers (== `push`/`pull_request`/`workflow_dispatch`) for upstream/forks
  OR
* on non-`pull_request` triggers (== `push`/`schedule`/`workflow_dispatch`) for upstream/forks

Which means it will still run the full set of four triggers for forks? 

image

---

This [discussion](https://github.com/orgs/community/discussions/38047) says:

> Currently, to disable running scheduled jobs for contributors, each job needs to be annotated with this.

```yml
if: github.event_name != 'schedule' || (github.event_name == 'schedule' && github.repository == 'arduino/arduino-cli')
```

So run:

* on any non-`schedule` triggers (== `push`/`pull_request`/`workflow_dispatch`) for upstream/forks
  OR
* on `schedule` triggers for upstream

Is that what we're after?

image

AlexWaygood:
Ah, good catch, I completely screwed up the logic here!

hugovk (approved):
Thank you!

PR #286, opened 2023-05-02
Add mypy config; run mypy in CI

AlexWaygood:
Work towards #282.

Adding type hints without validating them with a type checker is a bad idea, because the type hints can quickly go out of date with the source code they're describing. This PR adds some configuration for mypy, and a job to run mypy in CI.

Mypy is run on the `pyperformance/` directory, but `pyperformance/data-files/` and `pyperformance/tests/` are excluded from being typechecked. I've only enabled a few of the stricter settings for now, so that we can take advantage of "gradual typing" -- mypy will only emit errors on functions that have type annotations. It will assume all functions that don't have type annotations are meant to be ignored by the type checker.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,6 +68,8 @@ dependencies = [
 [project.optional-dependencies]
 dev = [
     'tox',
+    'mypy==1.2.0',
+    'tomli',  # Needed even on 3.11+ for typechecking with mypy
 ]
 
 [project.scripts]
@@ -81,3 +83,23 @@ find = {}  # Scanning implicit namespaces is active by default
 
 [tool.setuptools.dynamic]
 version = {attr = "pyperformance.__version__"}
+
+[tool.mypy]
+python_version = "3.7"
+pretty = true
+enable_error_code = "ignore-without-code"
+disallow_any_generics = true
+warn_redundant_casts = true
+warn_unused_ignores = true
+warn_unused_configs = true
+files = [
+    'pyperformance/',
+]
+exclude = [
+    'pyperformance/data-files/',
+    'pyperformance/tests/'
+]
+
+[[tool.mypy.overrides]]
+module = "pyperf"
+ignore_missing_imports = true
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -1,0 +1,29 @@
+name: mypy
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+env:
+  FORCE_COLOR: 1
+  TERM: xterm-256color # needed for FORCE_COLOR to work on mypy on Ubuntu, see https://github.com/python/mypy/issues/13817
+
+jobs:
+  mypy:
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -81,3 +83,23 @@ find = {}  # Scanning implicit namespaces is active by default
 
 [tool.setuptools.dynamic]
 version = {attr = "pyperformance.__version__"}
+
+[tool.mypy]
+python_version = "3.7"
```
AlexWaygood, line 88: It's generally a good idea to run mypy on the lowest Python version you support; this means it will flag if you accidentally use a feature that only exists in a newer Python version

pyproject.toml
```diff
@@ -81,3 +83,23 @@ find = {}  # Scanning implicit namespaces is active by default
...
+pretty = true
+enable_error_code = "ignore-without-code"
```
AlexWaygood, line 90: This means that mypy will emit an error if you do a bare `# type: ignore` without a specific error code (e.g. `# type: ignore[import]` for import-related errors)

pyproject.toml
```diff
@@ -81,3 +83,23 @@ find = {}  # Scanning implicit namespaces is active by default
...
+disallow_any_generics = true
```
AlexWaygood, line 91: This means that mypy will forbid using, for example `list` as an annotation. You should use `list[int]` instead, which gives the type checker much more information to work with.

pyproject.toml
```diff
@@ -81,3 +83,23 @@ find = {}  # Scanning implicit namespaces is active by default
...
+    'pyperformance/',
+]
+exclude = [
+    'pyperformance/data-files/',
+    'pyperformance/tests/'
+]
+
+[[tool.mypy.overrides]]
+module = "pyperf"
+ignore_missing_imports = true
```
AlexWaygood, line 105: pyperf doesn't have type annotations, so without this section, mypy would emit an error at every location pyperf was imported, which was quite annoying.

hauntsaninja (commented):
In case it's useful, this is the gradient of options I recommend: https://mypy.readthedocs.io/en/latest/existing_code.html#introduce-stricter-options

In particular, I'd add `strict_equality=True` and as you already know, work asap to getting `check_untyped_defs=True` passing.

AlexWaygood:
> In case it's useful, this is the gradient of options I recommend: https://mypy.readthedocs.io/en/latest/existing_code.html#introduce-stricter-options
> 
> In particular, I'd add `strict_equality=True` and as you already know, work asap to getting `check_untyped_defs=True` passing.

Thanks! My basic plan here is:

1. Add mypy config and a very basic CI check
2. Add as many annotations as possible through several PRs, making use of gradual typing while the codebase is still largely untyped
3. Once the codebase is largely typed, gradually increase mypy strictness

So I'm _slightly_ loathe to add any stricter settings right now that could flag pre-existing errors in the code, rather than errors related to the type annotations I'm adding (I'd rather deal with _that_ category of errors in stage 3). But `strict_equality=True` will probably be one of the first I'll add after stages 1 and 2 are complete!

PR #288, opened 2023-05-07
support --inherit-environ when reusing a venv

carljm:
The `--inherit-environ` flag was only inconsistently supported; it had no effect in case we were reusing an existing venv.


AlexWaygood (approved):
LGTM

Issue #287, opened 2023-05-07
asyncio gather vs TaskGroup in async_tree benchmarks


itamaro:
currently, the async_tree set of benchmarks uses `asyncio.gather`, but `asyncio.TaskGroup` is the newer cooler thing.

recently I worked on a few asyncio optimizations, and had to patch pyperformance (https://github.com/itamaro/pyperformance/commit/fe365c8c3b6020eb9918530d06133cb19f5335b2) to measure the impact on `TaskGroup` (in addition to the proposed patch in gh-279 to cover eager task execution).

as an aside, I was also able to compare gather vs TaskGroup (by comparing the results with and without that patch), and found that TaskGroup is faster than gather across the board!

I was wondering what would be the best way to address this in the benchmarks suite.
1. add yet another set of async_tree flavors that use TaskGroup (in addition to the ones using gather)? (considering the proposed gh-279, it's growing the matrix quite a bit)
2. change the existing set to use TaskGroup instead of gather? (which would make it 3.11+ only, and make it less useful when comparing to older runs that used gather)
3. keep the existing set as is, but use TaskGroup if it's available, falling back to gather otherwise?
4. don't bother - leave it with gather

itamaro:
@willingc @gvanrossum what do you think?

gvanrossum:
I think it's up to @mdboom. In general fixing benchmarks that already exist is painful enough that we don't do it unless there's no other way (e.g. a feature or dependency becomes obsolete). I am fine with adding another benchmark in the asyncio tree family. Maybe @kumaraditya303 has an opinion too?

kumaraditya303:
I vote for 1, we need benchmarks to cover both gather and taskgroups. Keep me in the loop for any PRs for this.

PR #293, opened 2023-05-27
gh-287: Add TaskGroups variants to async_tree benchmarks

itamaro:
This PR adds 8 new async_tree benchmark variants covering TaskGroups usage for the 8 original variants.
In addition, I also added asyncio tags to all these benchmarks, and added a requires-python field to restrict the TaskGroups variants to >=3.11 and the eager variants to >=3.12.

fixed gh-287

### DIFF
--- a/pyperformance/data-files/benchmarks/bm_async_tree/run_benchmark.py
+++ b/pyperformance/data-files/benchmarks/bm_async_tree/run_benchmark.py
@@ -58,19 +58,19 @@ class AsyncTree:
               for _ in range(NUM_RECURSE_BRANCHES)]
         )
 
-    async def recurse_with_task_groups(self, recurse_level):
+    async def recurse_with_task_group(self, recurse_level):
         if recurse_level == 0:
             await self.workload_func()
             return
 
-        await asyncio.gather(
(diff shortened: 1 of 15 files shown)


itamaro:
cc @kumaraditya303 you were interested in this

pyperformance/data-files/benchmarks/bm_async_tree/run_benchmark.py
```diff
@@ -47,17 +48,31 @@ async def workload_func(self):
...
+            *[self.recurse_with_gather(recurse_level - 1)
+              for _ in range(NUM_RECURSE_BRANCHES)]
+        )
+
+    async def recurse_with_task_groups(self, recurse_level):
+        if recurse_level == 0:
+            await self.workload_func()
+            return
+
+        await asyncio.gather(
```
kumaraditya303, line 66: This should use taskgroup right not gather?

itamaro: haha, of course, good catch! I used gather in both for local testing with 3.10 and forgot to change back 😬

kumaraditya303 (approved):

PR #296, opened 2023-05-30
Only run dependabot the "main" requirements.txt

mdboom:
You can't exclude directories for dependabot, only exclude them, (thanks @hugovk for discovering that), so by moving the main requirements.txt to a directory that doesn't include all of the benchmarks' requirements.txt beneath it, we can limit dependabot to just that one file.


hugovk (approved):

hugovk:
Please see https://github.com/python/pyperformance/pull/298 as a followup.

PR #300, opened 2023-06-02
Publish to PyPI using a trusted publisher

hugovk:
This means we don't need a long-lived token stored in this repo:

* https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/
* https://docs.pypi.org/trusted-publishers/using-a-publisher/

Needs an admin to log in to https://pypi.org/manage/project/pyperformance/settings/publishing/ and add a new publisher:

After that, we can delete `PYPI_API_TOKEN` from this repo's secrets and also the corresponding one at https://pypi.org/manage/account/ for whoever set it up.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,6 +10,10 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
 
+    permissions:
+      # IMPORTANT: this permission is mandatory for trusted publishing
+      id-token: write
+
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


pablogsal:
Done! Can someone check that it works?

pablogsal:
Screenshot 2023-06-13 at 16 27 48

hugovk:
Thanks!

Let's merge this and make a new PR to check it.

corona10 (approved):
lgtm

PR #303, opened 2023-06-16
Add a pure python sql engine benchmark.

tobymao:


mdboom (changes requested):
This is great, and thanks for the contribution.  One concern I have is that this may have an impact on the existing sqlglot benchmarks, and thus make the results no longer comparable with already-collected data.  We should see if those are affected in a statistically significant way, and if so, add this as a separate benchmark suite instead.

tobymao:
@mdboom how do you check that? the biggest change that would impact previous versions is that i upgraded sqlglot. but in terms of this test in isolation, i think the only impact may be in GC because it's a new test?

ambv:
Yeah, the problem is exactly that you upgraded sqlglot over several versions.

tobymao:
> Yeah, the problem is exactly that you upgraded sqlglot over several versions.

can't you just back test it and recompute the benchmarks with the new sqlglot version? the old versions of sqlglot don't have a sql engine

PR #321, opened 2023-10-22
Prepare release 1.10.0

corona10:


pyperformance/__init__.py
```diff
@@ -2,7 +2,7 @@
 import sys
 
 
-VERSION = (1, 0, 9)
+VERSION = (1, 10, 0)
```
diegorusso, line 5: Just out of curiosity. What made you jump from 0 to 10 in the minor number?

corona10: woo.. it was my mistake.

diegorusso: ok, so it should have been 1.0.10 then :) What are you going to do? Do you leave as it is or will you withdraw the release and re-release it with the fixed version?

corona10: Unfortunately, it was already released so we decided to keep the current version.
Sorry for the confusion :(

diegorusso: I'm fine with that! thanks for the update!

PR #326, opened 2024-01-26
Move jobs to correct location in benchmark.conf.sample

Fidget-Spinner:
Currently it's at the wrong location and doesn't actually do anything.


corona10:
https://github.com/python/pyperformance/blob/32c6bbf2b101a15a9b6bd555b5c12314bae2c232/pyperformance/commands.py#L203

Nice catch!

corona10 (approved):

PR #327, opened 2024-01-30
Add a feature for using the same number of loops as a previous run

mdboom:
**Motivation:**

On the Faster CPython team, we often collect pystats (counters of various interpreter events) by running the benchmark suite. It is very useful to compare the stats between two commits to see how a pull request affects the interpreter. Unfortunately, with pyperformance's default behavior where the number of loops is automatically calibrated, each benchmark may not be run the same number of times from run-to-run, making the data hard to compare.

This change adds a new argument to the `run` command which will use the same number of loops as a previous run. The `loops` for each benchmark is looked up from the metadata in the .json output of that previous run, and passed to the underlying call to `pyperf` using the `--loops` argument.

Additionally, this modifies one of the benchmarks, `sqlglot` to be compatible with that scheme. `sqlglot` is the only `run_benchmark.py` script that runs multiple benchmarks within it in a single call to the script.  This makes it impossible to set the number of loops independently for each of these benchmarks.  It's been updated to use the pattern from other "suites" of benchmarks (e.g. `async_tree`) where each benchmark has its own `.toml` file and is run independently.  This should still be backward compatible with older data collected from this benchmark, but doing `pyperformance run -b sqlglot` will now only run a single benchmark.

### DIFF
--- a/pyperformance/run.py
+++ b/pyperformance/run.py
@@ -68,7 +68,7 @@ def get_loops_from_file(filename):
 
 
 def run_benchmarks(should_run, python, options):
-    if hasattr(options, "same_loops"):
+    if options.same_loops is not None:
         loops = get_loops_from_file(options.same_loops)
     else:
         loops = {}
--- a/pyperformance/run.py
+++ b/pyperformance/run.py
@@ -1,5 +1,6 @@
 from collections import namedtuple
 import hashlib
+import json
 import sys
 import time
 import traceback
@@ -50,7 +51,28 @@ def get_run_id(python, bench=None):
     return RunID(py_id, compat_id, bench, ts)
 
 
+def get_loops_from_file(filename):
+    with open(filename) as fd:
+        data = json.load(fd)
+
+    loops = {}
+    for benchmark in data["benchmarks"]:
+        metadata = benchmark.get("metadata", data["metadata"])
+        name = metadata["name"]
+        if name.endswith("_none"):
+            name = name[:-len("_none")]
+        if "loops" in metadata:
+            loops[name] = metadata["loops"]
+
+    return loops
+
+
 def run_benchmarks(should_run, python, options):
+    if hasattr(options, "same_loops"):
+        loops = get_loops_from_file(options.same_loops)
+    else:
+        loops = {}
+
     to_run = sorted(should_run)
 
     info = _pythoninfo.get_info(python)
@@ -135,6 +157,10 @@ def run_benchmarks(should_run, python, options):
                     dest_suite = pyperf.BenchmarkSuite([res])
 
             return dest_suite
+
+        if name in loops:
+            pyperf_opts.append(f"--loops={loops[name]}")
+            pyperf_opts.append("--warmups=1")
 
         bench_venv, bench_runid = benchmarks.get(bench)
         if bench_venv is None:
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -1,6 +1,9 @@
 Changelog
 =========
 
+* Add a --same-loops option to the run command to use the exact same number of
+  loops as a previous run (without recalibrating).
+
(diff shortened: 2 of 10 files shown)


pyperformance/run.py
```diff
@@ -50,7 +51,28 @@ def get_run_id(python, bench=None):
     return RunID(py_id, compat_id, bench, ts)
 
 
+def get_loops_from_file(filename):
```
Fidget-Spinner, line 54: I'm confused. Is this a documented pyperperformance json format? Also it would be nice if we could add this to the `.conf` parser as well.

Fidget-Spinner (commented):
Looks good in general. I'll let Eric approve though.

pyperformance/run.py
mdboom: I don't know if it's documented, but it's the `.json` that's produced from `pyperformance run --output output.json`.

mdboom: That's a good suggestion about the `.conf` parser -- I never use it, so forgot about it.  But I'll go ahead and add support there, too.

Fidget-Spinner: Oh I see. Yeah that `.json` file is definitely documented then.

pyperformance/run.py
```diff
@@ -50,7 +51,28 @@ def get_run_id(python, bench=None):
...
+        data = json.load(fd)
+
+    loops = {}
+    for benchmark in data["benchmarks"]:
+        metadata = benchmark.get("metadata", data["metadata"])
+        name = metadata["name"]
+        if name.endswith("_none"):
+            name = name[:-len("_none")]
+        if "loops" in metadata:
+            loops[name] = metadata["loops"]
```
ericsnowcurrently, line 65: Would it be worth adding any checks for an invalid input file?  Probably not.  I expect any raised exceptions would already be clear enough.

pyperformance/run.py
```diff
@@ -136,6 +158,10 @@ def add_bench(dest_suite, obj):
 
             return dest_suite
 
+        if name in loops:
+            pyperf_opts.append(f"--loops={loops[name]}")
+            pyperf_opts.append("--warmups=1")
```
ericsnowcurrently, line 163: Hard-coding this to `1` feels off.  IIRC, the number of warmups can impact results on PyPy.  Is it something we can also extract from the `--same-loops` JSON file?  If you're sure `1` is correct then I'm okay with that.

ericsnowcurrently (approved):
LGTM

I left a couple of minor comments that you can take or leave.

It might be worth splitting out the bm_sqlglot changes into a separate PR, but it isn't a big deal.  I'll leave that up to you.

pyperformance/run.py
mdboom: Yeah, that's a good point.

mdboom: Looking at the data, it actually seems like we don't need this.

It looks like the first "run" of a benchmark has a number of warmups related to calibration -- as in it tries a bunch of times with an increasing loop number until it arrives at enough time:

```
          "warmups": [
            [1, 0.003556258001481183],
            [2, 0.0029844559976481833],
            [4, 0.0030109654981060885],
            [8, 0.003761511250559124],
            [16, 0.0030481524372589774],
            [32, 0.002865061812372005],
            [64, 0.0028616962342766783],
            [64, 0.002973547281271749],
            [64, 0.0028825146093822696],
            [64, 0.002945869875020435]
          ]
        },
```

These aren't really "warmups", they are calibration runs, but the data makes no distinction.

All subsequent runs just have one warmup:

```
          "warmups": [[64, 0.0028776703593393904]]
```

When you pass the `--loops` argument to `pyperf`, the whole "multiple warmups on the first run" thing goes away because it isn't doing calibration.

So, in short, I think I'll just take this line out.

PR #328, opened 2024-02-07
Add `list.sort` benchmarks

sobolevn:
As requested in https://github.com/python/cpython/pull/114687 by @corona10 I ported this script into `pyperformance`.

Changes:
1. Now we have new `sorting` group
2. All previous cases were translated into a separate benchmark (they are different only by the input data)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -14,6 +14,7 @@ Georg Brandl <georg@python.org>
 James Abbatiello <abbeyj@gmail.com>
 Jeffrey Yasskin <jyasskin@gmail.com>
 Maciej Fijalkowski
+Nikita Sobolev <mail@sobolevn.me>
 Reid Klecker <mk@mit.edu>
 Robert Grimm
 Skip Montanaro
--- a/doc/benchmarks.rst
+++ b/doc/benchmarks.rst
@@ -24,6 +24,7 @@ Available benchmark groups:
 * ``startup``: Collection of microbenchmarks focused on Python interpreter
   start-up time.
 * ``template``: Templating libraries
+* ``sorting``: Different sorting algorithms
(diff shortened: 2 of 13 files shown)


corona10:
@gvanrossum Do you think that it is okay to relocate `list.sort` benchmark in here?

sobolevn:
This still happens:

```
[40/88] list_sort_ascending_exchanged...
# /home/runner/work/pyperformance/pyperformance/venv/cpython3.11-3c206e65878c-compat-3c0fde653f89/bin/python -u /home/runner/work/pyperformance/pyperformance/pyperformance/data-files/benchmarks/bm_list_sort/run_benchmark.py list_sort_ascending_exchanged --debug-single-value --output /tmp/tmphqhu0yau --inherit-environ PYPERFORMANCE_RUNID
.
WARNING: the benchmark result may be unstable
* the shortest raw value is only 285 us

Try to rerun the benchmark with more runs, values and/or loops.
Run 'python -m pyperf system tune' command to reduce the system jitter.
Use pyperf stats, pyperf dump and pyperf hist to analyze results.
Use --quiet option to hide these warnings.
```

sobolevn:
`WARNING: the benchmark result may be unstable` is gone!

gvanrossum:
> @gvanrossum Do you think that it is okay to relocate `list.sort` benchmark in here?

My response is the same as for the decimal benchmark: moving this benchmark here is not going to help us speed up the Python interpreter, it will just make it harder for us to show any improvements. Maybe put this benchmark in Tools/scripts?

sobolevn:
> Maybe put this benchmark in Tools/scripts?

Sounds good to me! 👍

PR #329, opened 2024-03-05
Fix the django_template benchmark

mdboom:
This is broken by the removal of the `cgi` module in Python 3.13.  This adds the `legacy-cgi` PyPI library as a dependency as a workaround.


ericsnowcurrently (approved):
LGTM

PR #333, opened 2024-04-12
Fix mentioned Python versions to headline

HelgeCPH:
In the previous version, the document gave an example on how to compare Python 3.6 and Python 3.7 performance. However, the commands in the example referred partially to Python 3.8. Since this is confusing and not inline with the first line of the section, I changed the example so that it fits the text.


mdboom (approved):
LGTM

corona10 (approved):

PR #334, opened 2024-05-09
Skip dask benchmark on Windows for Python 3.13

mdboom:
The [dask benchmark hangs indefinitely on Python 3.13 on Windows](https://github.com/python/cpython/issues/118715).  This will just raise an exception instead to fix the benchmark.

There is already [a fix for this bug in dask here](https://github.com/dask/distributed/commit/e4a054505f1d584d920d6d1df2c419074de45d05), however, until that is released, we should guard against it so the benchmark suite doesn't hang on Windows.


corona10 (approved):
lgtm!

mdboom:
Thanks, @corona10.  Can you merge this?

PR #339, opened 2024-05-19
Benchmark a simple BPE tokeniser

hauntsaninja:
This isn't fully realistic, e.g. there are some optimisations you'd do if you actually cared about performance.
This benchmark currently takes about 3.5s, let me know if I should adjust that.


pyperformance/data-files/benchmarks/bm_bpe_tokeniser/run_benchmark.py
```diff
@@ -0,0 +1,156 @@
+"""
+Benchmark a BPE tokeniser.
+
+Author: Shantanu Jain
+
+Based on code from tiktoken.
```
mdboom, line 6: Is there a repo or source we should link to for credit?

mdboom (approved):
This looks great.  Thanks for the submission.

I had just the one comment in case there is a source to link to.

hauntsaninja:
@mdboom are you around in the sprints? I'd love to chat about what more good benchmarks could be

PR #341, opened 2024-06-14
Add benchmarks for the decimal module

mdboom:
These currently exist in the [CPython source](https://github.com/python/cpython/blob/main/Modules/_decimal/tests/bench.py), but it would be good to have them here for continuous benchmarking.


ericsnowcurrently (approved):
LGTM

Were you planning on adding the new benchmarks to the suite (i.e. pyperformance/data-files/benchmarks/MANIFEST)?  At least you could add them there and leave them commented out.  Is there any reason leave them out though?

PR #342, opened 2024-06-14
Add a YAML-parsing benchmark

mdboom:


pyperformance/data-files/benchmarks/bm_yaml/run_benchmark.py
```diff
@@ -0,0 +1,91 @@
+"""
+Script for testing the performance of YAML parsing, using yaml.
+
+This will dump/load several real world-representative objects a few thousand
+times. The methodology below was chosen for was chosen to be similar to
+real-world scenarios which operate on single objects at a time.
```
ericsnowcurrently, line 6:
```suggestion
This will dump/load several real world-representative objects a few thousand
times. The methodology below was chosen  to be similar to
real-world scenarios which operate on single objects at a time.
```

pyperformance/data-files/benchmarks/bm_yaml/run_benchmark.py
```diff
@@ -0,0 +1,91 @@
...
+
+This explicitly tests the pure Python implementation in pyyaml, not its C
+extension.
```
ericsnowcurrently, line 9: This is a helpful comment.  It may be worth moving it next to the line where we take this action.  I'm guessing that's the `Loader=yaml.Loader` bit on line 70.

pyperformance/data-files/benchmarks/bm_yaml/run_benchmark.py
```diff
@@ -0,0 +1,91 @@
...
+    'username': 'collinwinter',
+    'play_count': 9,
+    'view_count': 7,
+    'zip': ''}
+
+TUPLE = (
+    [265867233, 265868503, 265252341, 265243910, 265879514,
+     266219766, 266021701, 265843726, 265592821, 265246784,
+     265853180, 45526486, 265463699, 265848143, 265863062,
+     265392591, 265877490, 265823665, 265828884, 265753032], 60)
```
ericsnowcurrently, line 61: It may be worth it to also/instead use some real-world data, like we do for bm_tomli_loads?

ericsnowcurrently (approved):
LGTM

I've left a couple of very minor comments that I'll leave to you to resolve/ignore before merging.

PR #343, opened 2024-06-14
Add pypdf benchmark

mdboom:


pyperformance/data-files/benchmarks/bm_pypdf/run_benchmark.py
```diff
@@ -0,0 +1,89 @@
+"""
+A simple pypdf benchmark.
+
+Adapted from pypdf's own benchmarks:
+
+   https://github.com/py-pdf/pypdf/blob/main/tests/bench.py
```
ericsnowcurrently, line 6: Assuming it makes sense to do so, could we contribute a change upstream to make their benchmarks compatible with pyperformance (and pull them in like we do for the pyston benchmarks), instead of landing a new benchmark here?

pyperformance/data-files/benchmarks/bm_pypdf/run_benchmark.py
```diff
@@ -0,0 +1,89 @@
...
+    page.extract_text()
+
+
+def bench_page_ops(loops):
+    """
+    Apply various page operations.
+
+    Rotation, scaling, translation, content stream compression, text extraction
+    """
+    content = (DATA_DIR / "libreoffice-writer-password.pdf").read_bytes()
```
ericsnowcurrently, line 58: FWIW, GitHub says this is an invalid PDF file, though I'm sure it's fine.

Also, it may be worth adding a note about where this file came from and what's in it, especially since it's a binary format.

mdboom: Yeah, that's a good point.  It's a password-protected PDF file, which is probably why Github doesn't like it.  I just grabbed what pypdf was using -- we could choose something more deliberately.

Issue #345, opened 2024-06-18
Allow pyperformance to compile CPython with the JIT enabled.


diegorusso:
From 3.13, CPython can be built with the JIT and it doesn't seem to be the way to pass custom flags to the configure or the compile step: https://github.com/python/pyperformance/blob/784d0421f387f6245ecbbe744a530c238c60c741/pyperformance/compile.py#L277

It would be nice to have either:
* a way to pass custom flags to the configure via CLI/benchmark.conf
* hard code the jit (like lto/pgo) in the benchmark.conf

I don't have strong opinions for both implementations and TBH we could have both. The former is more generic and flexible, the latter more explicit.

brandtbucher:
@mdboom, would this be helpful for us?

mdboom:
I think it's a good idea in order to make it easier for others.  The Faster CPython team's [infrastructure](https://github.com/faster-cpython/benchmarking-public) doesn't actually use `pyperformance compile`, but in the interests of open source, it probably wouldn't be a bad idea to close the gap there so `pyperformance compile` is feature-complete and our infrastructure could start using it.

diegorusso:
Thanks for confirming that. I guess the only way possible for now is to compile Python outside of pyperformance and then instruct pyperformance to run benchmarks with the just compiled CPython.
I agree with @mdboom, I think it is useful if we can expose both the JIT (and maybe the free threading) as configuration options.
For the sake of flexibility it would be nice to expose an interface to pass whatever flags we want to the `./configure`.

diegorusso:
PR: https://github.com/python/pyperformance/pull/351

PR #351, opened 2024-08-22
Add support for just-in-time (JIT) builds

diegorusso:
This PR implements https://github.com/python/pyperformance/issues/345 by specifically having the `jit` option in the configuration file.

### DIFF
--- a/doc/benchmark.conf.sample
+++ b/doc/benchmark.conf.sample
@@ -47,8 +47,7 @@ pgo = True
 #   be used to enable it at runtime.
 # - interpreter: do not build the JIT, but do build and enable the micro-op
 #   interpreter. This is useful for those of us who find ourselves developing
-#   or debugging micro-ops (but don’t want to deal with the JIT), and is
-#   equivalent to using the -X uops or PYTHON_UOPS=1 options today.
+#   or debugging micro-ops (but don’t want to deal with the JIT).
 #   PYTHON_JIT=0 can be used to disable the micro-op interpreter at runtime.
 jit = no
 
--- a/doc/benchmark.conf.sample
+++ b/doc/benchmark.conf.sample
@@ -37,6 +37,21 @@ lto = True
 # Profiled Guided Optimization (PGO)?
 pgo = True
 
+# Build the experimental just-in-time (JIT) compiler?
+# Possible values are:
+# - no: (default) do not build the JIT or the micro-op interpreter.
+#   The new PYTHON_JIT environment variable has no effect.
+# - yes: build the JIT and enable it by default. PYTHON_JIT=0 can be used to
+#   disable it at runtime.
+# - yes-off: build the JIT, but do not enable it by default. PYTHON_JIT=1 can
+#   be used to enable it at runtime.
+# - interpreter: do not build the JIT, but do build and enable the micro-op
+#   interpreter. This is useful for those of us who find ourselves developing
+#   or debugging micro-ops (but don’t want to deal with the JIT), and is
+#   equivalent to using the -X uops or PYTHON_UOPS=1 options today.
+#   PYTHON_JIT=0 can be used to disable the micro-op interpreter at runtime.
+jit = no
+
 # The space-separated list of libraries that are package-only,
 # i.e., locally installed but not on header and library paths.
 # For each such library, determine the install path and add an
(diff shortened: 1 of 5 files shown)


diegorusso:
Tagging @brandtbucher and @mdboom for awareness.

doc/benchmark.conf.sample
```diff
@@ -37,6 +37,21 @@ lto = True
...
+# - no: (default) do not build the JIT or the micro-op interpreter.
+#   The new PYTHON_JIT environment variable has no effect.
+# - yes: build the JIT and enable it by default. PYTHON_JIT=0 can be used to
+#   disable it at runtime.
+# - yes-off: build the JIT, but do not enable it by default. PYTHON_JIT=1 can
+#   be used to enable it at runtime.
+# - interpreter: do not build the JIT, but do build and enable the micro-op
+#   interpreter. This is useful for those of us who find ourselves developing
+#   or debugging micro-ops (but don’t want to deal with the JIT), and is
+#   equivalent to using the -X uops or PYTHON_UOPS=1 options today.
```
brandtbucher, line 51: Outdated info:

```suggestion
#   or debugging micro-ops (but don’t want to deal with the JIT).
```

pyperformance/compile.py
```diff
@@ -291,6 +291,8 @@ def compile(self):
             config_args.append('--with-pydebug')
         elif self.conf.lto:
             config_args.append('--with-lto')
+        if self.conf.jit and self.conf.jit in ("yes", "yes-off", "interpreter"):
```
brandtbucher, line 294: Is there a reason you're choosing to validate this? I'd lean towards *not* validating it since any new options or defaults on the CPython side would need to be mirrored here.

```suggestion
        if self.conf.jit:
```

pyperformance/compile.py
```diff
@@ -801,6 +803,7 @@ def getint(section, key, default=None):
         conf.directory = getfile('compile', 'bench_dir')
         conf.lto = getboolean('compile', 'lto', True)
         conf.pgo = getboolean('compile', 'pgo', True)
+        conf.jit = getstr('compile', 'jit', 'no')
```
brandtbucher, line 806: Maybe default to the empty string, in case the upstream default gets changed at some point?

```suggestion
        conf.jit = getstr('compile', 'jit', '')
```

pyperformance/tests/data/find-pyperformance.py
```diff
@@ -13,7 +13,7 @@
 testsroot = os.path.dirname(datadir)
 pkgroot = os.path.dirname(testsroot)
 reporoot = os.path.realpath(os.path.dirname(pkgroot))
-marker = os.path.join(reporoot, 'setup.py')
+marker = os.path.join(reporoot, 'pyproject.toml')
```
brandtbucher, line 16: This seems unrelated?

tox.ini
```diff
@@ -13,7 +13,7 @@ basepython = python3
 basepython = python3
 deps=
     sphinx
-whitelist_externals = make
+allowlist_externals = make
```
brandtbucher, line 16: Ditto. Not a bad change, but just double-checking that it doesn't really have anything to do with the change in this PR.

brandtbucher (commented):
Thanks!

pyperformance/tests/data/find-pyperformance.py
diegorusso: I wanted to run tests and this was a blocker for running them. I left this change in a different commit so if the PR is merged without squashing this change and the next one will be in a separate commit.
If you want I can create a separate PR to include these to small changes.

diegorusso: This is the commit: https://github.com/python/pyperformance/pull/351/commits/f47ab7c5d0392cb8c71b53a9dc557076e1ab7bf7

tox.ini
diegorusso: This is the commit: https://github.com/python/pyperformance/pull/351/commits/f47ab7c5d0392cb8c71b53a9dc557076e1ab7bf7

pyperformance/tests/data/find-pyperformance.py
diegorusso: Dropped, I'll raise a new PR

tox.ini
diegorusso: Dropped, I'll raise a new PR

diegorusso:
@brandtbucher I moved the unrelated changes to this PR: https://github.com/python/pyperformance/pull/352 and I have addressed your feedback. Thanks!

brandtbucher:
Looks good to me, thanks!

Issue #353, opened 2024-09-13
Implement timeout mechanism


diegorusso:
Hello,

as discussed on [Discourse](https://discuss.python.org/t/add-timeout-management-in-pyperformance/63728) it would be nice to implement a timeout within pyperfomance.

If a benchmark run hangs, pyperformance doesn’t have a mechanism to time out and terminate the underlying process. We’ve had a few cases where a benchmark was hanging literally forever (deadlock) and the job was terminated by the CI system eventually (after many hours!). This is not ideal because:

* it’s hard to set an appropriate timeout for the whole process as it might vary from machine to machine
* sometimes folks might not have full control of the CI system hence impossible to set a sensible timeout
* having an early failure by pyperformance is preferred than having a timeout at process (pyperformance) level. This allow better usage of CI resources.

Requirements:
* Ideally the timeout would be applied at benchmark run level and not for the whole pyperformance process allowing a more fine grained control in case of timeout.
* If a benchmark times out, it should not fail the whole pyperformance suite but we should mark it as failed (there is already a mechanism to do that).
* There should be able to provide an option at cli level to override the default timeout value.

diegorusso:
I plan to implement this soon.

diegorusso:
This is now down in pyperf (https://github.com/psf/pyperf/pull/205) but before merging the PR in pyperformance we need a release in pyperf.

PR #354, opened 2024-09-19
Pass --timeout flag to pyperf

diegorusso:
Add the `--timeout` flag to the `run` command. By default there is no timeout and user needs to specify it if they want it. I wanted to avoid to change the current behaviour.
This addressed the issue https://github.com/python/pyperformance/issues/353

If you run the below command passing the timeout of 1 second, dask will fail and it will be reported at the end.

```
$ pyperformance run --debug-single-value -b float,go,mako,dask --timeout 1
...
...
Performance version: 1.11.0
Python version: 3.12.3 (64-bit)
Report on Linux-6.8.0-44-generic-aarch64-with-glibc2.39
Number of logical CPUs: 8
Start date: 2024-09-19 10:11:20.472768
End date: 2024-09-19 10:11:21.843698

### float ###
85.8 ms

### go ###
119 ms

### mako ###
9.77 ms

1 benchmarks failed:
- dask (Timed out after 1 seconds)
```

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -101,9 +101,8 @@ Usage::
 
   pyperformance run [-h] [-r] [-f] [--debug-single-value] [-v] [-m]
                        [--affinity CPU_LIST] [-o FILENAME]
-                       [--append FILENAME] [--min-time MIN_TIME]
-                       [--same-loops SAME_LOOPS] [--timeout TIMEOUT]
-                       [--manifest MANIFEST] [-b BM_LIST]
+                       [--append FILENAME] [--manifest MANIFEST]
+                       [--timeout TIMEOUT] [-b BM_LIST]
                        [--inherit-environ VAR_LIST] [-p PYTHON]
 
 options::
@@ -125,17 +124,12 @@ options::
                         baseline_python, not changed_python.
   --append FILENAME     Add runs to an existing file, or create it if
                         it doesn't exist
-  --min-time MIN_TIME   Minimum duration in seconds of a single value, used
-                        to calibrate the number of loops
-  --same-loops SAME_LOOPS
-                        Use the same number of loops as a previous run
-                        (i.e., don't recalibrate). Should be a path to a
-                        .json file from a previous run.
-  --timeout TIMEOUT     Timeout for a benchmark run (default: disabled)
+  --timeout TIMEOUT     Specify a timeout in seconds for a single
+                        benchmark run (default: disabled)
   --manifest MANIFEST   benchmark manifest file to use
   -b BM_LIST, --benchmarks BM_LIST
-                        Comma-separated list of benchmarks or groups to run.
-                        Can contain both positive and negative arguments:
+                        Comma-separated list of benchmarks to run. Can
+                        contain both positive and negative arguments:
                         --benchmarks=run_this,also_this,-not_this. If
                         there are no positive arguments, we'll run all
                         benchmarks except the negative arguments.
@@ -148,6 +142,10 @@ options::
   -p PYTHON, --python PYTHON
                         Python executable (default: use running
                         Python)
+  --same-loops SAME_LOOPS
+                        Use the same number of loops as a previous run
+                        (i.e., don't recalibrate). Should be a path to a
+                        .json file from a previous run.
 
 show
 ----
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -101,9 +101,10 @@ Usage::
 
   pyperformance run [-h] [-r] [-f] [--debug-single-value] [-v] [-m]
                        [--affinity CPU_LIST] [-o FILENAME]
-                       [--append FILENAME] [--manifest MANIFEST]
-                       [-b BM_LIST] [--inherit-environ VAR_LIST]
-                       [-p PYTHON]
+                       [--append FILENAME] [--min-time MIN_TIME]
+                       [--same-loops SAME_LOOPS] [--timeout TIMEOUT]
+                       [--manifest MANIFEST] [-b BM_LIST]
+                       [--inherit-environ VAR_LIST] [-p PYTHON]
 
 options::
 
@@ -124,10 +125,17 @@ options::
                         baseline_python, not changed_python.
   --append FILENAME     Add runs to an existing file, or create it if
                         it doesn't exist
+  --min-time MIN_TIME   Minimum duration in seconds of a single value, used
+                        to calibrate the number of loops
+  --same-loops SAME_LOOPS
+                        Use the same number of loops as a previous run
+                        (i.e., don't recalibrate). Should be a path to a
+                        .json file from a previous run.
+  --timeout TIMEOUT     Timeout for a benchmark run (default: disabled)
   --manifest MANIFEST   benchmark manifest file to use
   -b BM_LIST, --benchmarks BM_LIST
-                        Comma-separated list of benchmarks to run. Can
-                        contain both positive and negative arguments:
+                        Comma-separated list of benchmarks or groups to run.
+                        Can contain both positive and negative arguments:
                         --benchmarks=run_this,also_this,-not_this. If
                         there are no positive arguments, we'll run all
                         benchmarks except the negative arguments.
@@ -140,10 +148,6 @@ options::
   -p PYTHON, --python PYTHON
                         Python executable (default: use running
                         Python)
-  --same-loops SAME_LOOPS
-                        Use the same number of loops as a previous run
-                        (i.e., don't recalibrate). Should be a path to a
-                        .json file from a previous run.
 
 show
 ----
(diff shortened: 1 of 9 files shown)


nineteendo:
Could you document `--min-time` in a different pull request?

doc/usage.rst
```diff
@@ -101,9 +101,10 @@ Usage::
 
   pyperformance run [-h] [-r] [-f] [--debug-single-value] [-v] [-m]
                        [--affinity CPU_LIST] [-o FILENAME]
-                       [--append FILENAME] [--manifest MANIFEST]
-                       [-b BM_LIST] [--inherit-environ VAR_LIST]
-                       [-p PYTHON]
+                       [--append FILENAME] [--min-time MIN_TIME]
+                       [--same-loops SAME_LOOPS] [--timeout TIMEOUT]
```
nineteendo, line 105: Could you document `--min-time` and `--same-loops` in a different pull request?

pyperformance/commands.py
```diff
@@ -191,8 +191,8 @@ def cmd_run(options, benchmarks):
 
     if errors:
         print("%s benchmarks failed:" % len(errors))
-        for name in errors:
-            print("- %s" % name)
+        for name, reason in errors:
+            print("- %s (%s)" % (name, reason))
```
nineteendo, line 195: Should this be in a separate pull request?

doc/usage.rst
diegorusso: `--same-loops` has just moved from the end of the list to the right place. it was misplaced in the first place.
`--min-time` was never added here but it's not new.

Shall I put both of them in a PR?

vstinner (commented):
pyperf main process does basically nothing, it just waits until worker processes complete. It may be safer to implement the timeout in pyperf, to kill both processes, the main one and the current worker process.

If you implement the timeout in pyperformance, you only see the main process and you cannot (easily) kill the worker process.

doc/usage.rst
nineteendo: Moving it to a different place is probably fine, but updating the usage should probably be in a separate PR.

pyperformance/commands.py
diegorusso: The timeout implementation introduces a new way that a benchmark can fail. This is really to differentiate the ways a benchmarks can terminate hence I thought to be explicit and communicate it back to the user what the reason of the termination is.
If you want I can split it in an extra PR.

nineteendo: It's not up to me to decide that. I'm not a maintainer. :)

doc/usage.rst
mdboom: I implemented `--same-loops` a few months ago -- it's documented in here and in the usage.rst.  Was there something else you are looking for?  I'll volunteer to submit if so (in a separate PR).

doc/usage.rst
```diff
@@ -124,10 +125,17 @@ options::
                         baseline_python, not changed_python.
   --append FILENAME     Add runs to an existing file, or create it if
                         it doesn't exist
+  --min-time MIN_TIME   Minimum duration in seconds of a single value, used
+                        to calibrate the number of loops
+  --same-loops SAME_LOOPS
+                        Use the same number of loops as a previous run
+                        (i.e., don't recalibrate). Should be a path to a
+                        .json file from a previous run.
+  --timeout TIMEOUT     Timeout for a benchmark run (default: disabled)
```
mdboom, line 134: What's the unit here?  Seconds?

pyperformance/commands.py
mdboom: This change to show the reason makes sense to me.

pyperformance/tests/__init__.py
```diff
@@ -39,6 +39,10 @@ def run_cmd(cmd, *args, capture=None, onfail='exit', verbose=True):
 
     if verbose:
         print(f"(tests) Execute: {argv_str}", flush=True)
+
+    if timeout:
+        kwargs['timeout'] = 60
```
mdboom, line 44: Should we add a test that deliberately times out to confirm it's working?

mdboom (commented):
I may have sent you forth with bad information by suggesting that this would be easier in `pyperformance`, and maybe @vstinner is right that this would be better handled in `pyperf`.

We'd still need a patch to `pyperformance` to pass the `--timeout` value down to pyperf -- 80% of this PR is handling commandline arguments as it is, so not wasted effort.

doc/usage.rst
diegorusso: yes, they are seconds. I think I should be explicit. I'll address it tomorrow.

doc/usage.rst
diegorusso: The changes in usage.rst are there to reflect the actual output of pyperformance. In the PR https://github.com/python/pyperformance/pull/327 implemented the same loops option but it has been forgotten to be added here.

doc/usage.rst
```diff
@@ -124,10 +125,17 @@ options::
                         baseline_python, not changed_python.
   --append FILENAME     Add runs to an existing file, or create it if
                         it doesn't exist
+  --min-time MIN_TIME   Minimum duration in seconds of a single value, used
+                        to calibrate the number of loops
+  --same-loops SAME_LOOPS
```
diegorusso, line 130: Similarly to the comment above (re. same loops), argparse positions the option here and not at the end like used to be. This because it is an option of the run command.
--manifest, -b and -p flags are option of pyperformance hence they are placed at the end.

pyperformance/tests/__init__.py
diegorusso: yes, it makes sense. I'll add it tomorrow.

diegorusso:
> I may have sent you forth with bad information by suggesting that this would be easier in `pyperformance`, and maybe @vstinner is right that this would be better handled in `pyperf`.
> 
> We'd still need a patch to `pyperformance` to pass the `--timeout` value down to pyperf -- 80% of this PR is handling commandline arguments as it is, so not wasted effort.

Don't worry, that's the purpose of the review. I leave this PR open for now and implement the timeout in pyperf. Once that one is in, I'll fix this up to make use of the timeout.

doc/usage.rst
diegorusso: Dropped from this PR

doc/usage.rst
diegorusso: Added.

diegorusso:
I've addressed feedbacks on this PR and also raised a new one in pyperf: https://github.com/psf/pyperf/pull/205

pyperformance/tests/__init__.py
diegorusso: As the timeout is implemented in pyperf now (https://github.com/psf/pyperf/pull/205) I created a test in that PR. pyperformance now just passes the info to pyperf.
Would you like to have a test that checks the timeout is actually passed to pyperf?

mdboom: My comment was for earlier when timeout was handled here. I'm fine with not having a timeout test in pyperformance now that the actual implementation is in pyperf.

diegorusso:
Just to let you know that the [pyperf PR](https://github.com/psf/pyperf/pull/205) has been merged (thanks @vstinner) so we have the timeout functionality there.

vstinner:
> Just to let you know that the https://github.com/psf/pyperf/pull/205 has been merged (thanks @vstinner) so we have the timeout functionality there.

You need a pyperf release first.

mdboom:
> > Just to let you know that the [psf/pyperf#205](https://github.com/psf/pyperf/pull/205) has been merged (thanks @vstinner) so we have the timeout functionality there.
> 
> You need a pyperf release first.

I started one [here](https://github.com/psf/pyperf/pull/206)

vstinner:
Your PR should update pyperf to 2.8.

diegorusso:
> Your PR should update pyperf to 2.8.

I see that pyperf is unpinned so in theory I should not change anything in there. I have done a fresh install of pyperformance and in fact it installs pyperf 2.8

mdboom:
> > Your PR should update pyperf to 2.8.
> 
> I see that pyperf is unpinned so in theory I should not change anything in there. I have done a fresh install of pyperformance and in fact it installs pyperf 2.8

I think you still need to update it [here](https://github.com/python/pyperformance/blob/main/pyperformance/requirements/requirements.txt#L13), which is the version that will actually run in the venvs.

diegorusso:
> > > Your PR should update pyperf to 2.8.
> > 
> > 
> > I see that pyperf is unpinned so in theory I should not change anything in there. I have done a fresh install of pyperformance and in fact it installs pyperf 2.8
> 
> I think you still need to update it [here](https://github.com/python/pyperformance/blob/main/pyperformance/requirements/requirements.txt#L13), which is the version that will actually run in the venvs.

Yes, you are right. I've just found it! :)

diegorusso:
I'm currently running the tests because there is a small change in the way pyperf reports unit measure: https://github.com/psf/pyperf/commit/432c419cff97804de8c780e82d2a8562e977275a

vstinner (approved):
LGTM

mdboom (approved):
Thanks!

PR #355, opened 2024-09-23
Drop support for Python 3.7 and 3.8

hugovk:
Python 3.8 is end-of-life next month:

* https://devguide.python.org/versions/
* https://peps.python.org/pep-0569/#lifespan

Many of the files under `pyperformance/data-files/benchmarks/` have `requires-python = ">=3.7"` or `requires-python = ">=3.8"` but I've not changed them.


mdboom (approved):
LGTM.

I agree with your policy of not changing the minimum versions on benchmarks, unless we change their content which should be exceedingly rare.

Issue #356, opened 2024-09-27
Base manifest doesn't allow  benchmark groups except "default" benchmark group


guywilsonjr:
Where applicable, please fill out the details below to help us troubleshoot the
issue that you are facing. :cake:

**How to reproduce the issue**
Using script:
```
python3 -m venv testvenv
source testvenv/bin/activate
pip install pyperformance
testvenv/bin/pyperformance run -b math
```

**What you expected to happen**
The tool will run all the math benchmarks producing output like: 

```
[1/3] float...
.....................
float: Mean +- std dev: 48.3 ms +- 1.6 ms
[2/3] nbody...
.....................
nbody: Mean +- std dev: 46.6 ms +- 1.7 ms
[3/3] pidigits...

```

**What actually happens**

```
Collecting pyperformance
  Using cached pyperformance-1.11.0-py3-none-any.whl.metadata (2.0 kB)
Collecting pyperf (from pyperformance)
  Using cached pyperf-2.7.0-py3-none-any.whl.metadata (6.0 kB)
Collecting packaging (from pyperformance)
  Using cached packaging-24.1-py3-none-any.whl.metadata (3.2 kB)
Collecting psutil>=5.9.0 (from pyperf->pyperformance)
  Using cached psutil-6.0.0-cp36-abi3-manylinux_2_12_x86_64.manylinux2010_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (21 kB)
Using cached pyperformance-1.11.0-py3-none-any.whl (4.2 MB)
Using cached packaging-24.1-py3-none-any.whl (53 kB)
Using cached pyperf-2.7.0-py3-none-any.whl (139 kB)
Using cached psutil-6.0.0-cp36-abi3-manylinux_2_12_x86_64.manylinux2010_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl (290 kB)
Installing collected packages: psutil, packaging, pyperf, pyperformance
Successfully installed packaging-24.1 psutil-6.0.0 pyperf-2.7.0 pyperformance-1.11.0
WARNING:root:no benchmark named 'math'
Python benchmark suite 1.11.0



ERROR: No benchmark was run
```

guywilsonjr:
I've create a PR to fix this: https://github.com/python/pyperformance/pull/357

PR #357, opened 2024-09-27
Update MANIFEST

guywilsonjr:
According to https://pyperformance.readthedocs.io/benchmarks.html#available-groups

As it stands by using the default manifest these groups are disabled. This PR is to re-enable their selection. I've tested this locally and it looks like it's working


guywilsonjr:
This should address https://github.com/python/pyperformance/issues/356

mdboom (approved):
Sorry for the delay in getting to this.  LGTM.

PR #358, opened 2024-09-27
Generate digital attestations for PyPI (PEP 740)

hugovk:
PEP 740 ("Index support for digital attestations") introduces signatures which links the PyPI package to the GitHub repo, and helps users verify the source and authenticity of packages.

PyPI is still implementing support, but we can already start using it, which should also help them test out.

* https://peps.python.org/pep-0740/
* https://github.com/pypa/gh-action-pypi-publish#generating-and-uploading-attestations
* https://github.com/pypi/warehouse/issues/15871


mdboom (approved):
I don't really know much about the attestations, so I was waiting someone else would come along and review.  But this really doesn't seem at all risky, so I'm going to merge.

PR #360, opened 2024-10-02
Fix #359: Add a Sphinx benchmark

### LINKED ISSUE #359: Add a Sphinx benchmark
The regression in [Sphinx caused by the incremental gc changes in CPython](https://github.com/python/cpython/issues/124567) was largely missed by the pyperformance benchmarking suite.  We should add a benchmark that reproduces that case.

mdboom:
This is mainly aimed at making sure our benchmarks catch whatever caused [the GC regression in CPython](https://github.com/python/cpython/issues/124567).  I measured about a 10% speedup on this when reverting the GC changes (and about 50% with debug builds, but performance of debug builds is not guaranteed anyway).


pyperformance/data-files/benchmarks/bm_sphinx/run_benchmark.py
```diff
@@ -0,0 +1,137 @@
...
+        "-j",
+        "1",
+        "-Q",
+        "--fresh-env",
+        "--write-all",
+        str(doc_root),
+        str(doc_root / "build" / "html"),
+    ]
+    t0 = pyperf.perf_counter()
+    sphinx_main()
```
AA-Turner, line 106: Using the `dummy` builder avoids writing HTML, and the slowdown / creation of reference cycles is in the parsing (reading) step rather than writing. 

Also this passes arguments directly rather than patching `sys.argv`

```suggestion
    t0 = pyperf.perf_counter()
    sphinx_main([
        "--builder",
        "dummy",
        "--doctree-dir",
        str(doc_root / "build" / "doctrees"),
        "--jobs",
        "1",
        "--silent",
        "--fresh-env",
        "--write-all",
        str(doc_root),
        str(doc_root / "build" / "html"),
    ])
```

pyperformance/data-files/benchmarks/bm_sphinx/data/Doc/tools/check-warnings.py
AA-Turner, line 1: This file can be deleted, it isn't used in the build.

mdboom:
Thanks for the feedback, @AA-Turner.  I think I've addressed them all (and fixed the tests).

AA-Turner:
I've created a series of commits to potentially reduce the size of the benchmark -- and opened https://github.com/mdboom/pyperformance/pull/1 for easy viewing of them.

You may not want to include some (or any) of them, but the idea was to remove irrelevant things from the benchmark data (& also hopefully take preventative measures against so that the `pyperformance` repo doesn't grow in size *too* much!)

A

pyperformance/data-files/benchmarks/bm_sphinx/requirements.txt
```diff
@@ -0,0 +1,2 @@
+sphinx==6.2.1
```
AA-Turner, line 1: Any reason to use 6.2? The CPython docs (currently) use 8.0.2, which is the version in which the regression was noticed.

mdboom: It doesn't require installing a separate, newer version of docutils just for this benchmark (saves some time).  I picked it because it's the current minimum required version for CPython, but that's kind of arbitrary.

AA-Turner: I might suggest `Sphinx==7.3.7` then, it supports Docutils 18.1 (used in `bm_docutils`) and also the longer and less cryptic command line options I suggested above.

However as you say the choice is somewhat arbitrary and more important to have the benchmark.

A

mdboom: Thanks.  That makes sense.

pyperformance/data-files/benchmarks/bm_sphinx/run_benchmark.py
```diff
@@ -0,0 +1,127 @@
...
+    t0 = pyperf.perf_counter()
+    sphinx_main(
+        [
+            "-b",
+            "dummy",
+            "-d",
+            str(doc_root / "build" / "doctrees"),
+            "-j",
+            "1",
+            "-Q",
```
AA-Turner, line 95:
```suggestion
            "--builder",
            "dummy",
            "--doctree-dir",
            str(doc_root / "build" / "doctrees"),
            "--jobs",
            "1",
            "--silent",
```

AA-Turner (approved):
Thanks! 

A

PR #361, opened 2024-10-09
Support specifying hooks in pyperformance

mdboom:
This is follow-on work to [supporting hooks in pyperf](https://github.com/psf/pyperf/pull/193), and simply allows specifying hooks to run from pyperformance.


ericsnowcurrently (approved):
LGTM

(caveat: I don't know much about pyperf hooks)

Would it be worth adding a test?

ericsnowcurrently (approved):
LGTM

PR #362, opened 2024-10-15
Fix a docstring in async_tree benchmark

mdboom:
Since https://github.com/python/cpython/pull/124899, argparse help strings are more strict, and will fail with invalid `%` format strings.


ericsnowcurrently (approved):
LGTM

PR #366, opened 2024-10-25
Make Windows a non-experimental platform in CI

mdboom:
This also fixes various things about the benchmarks so the unit tests pass on Windows.

### DIFF
--- a/pyperformance/data-files/benchmarks/bm_asyncio_websockets/run_benchmark.py
+++ b/pyperformance/data-files/benchmarks/bm_asyncio_websockets/run_benchmark.py
@@ -8,6 +8,7 @@ Author: Kumar Aditya
 import pyperf
 import websockets.server
 import websockets.client
+import websockets.exceptions
 import asyncio
 
 CHUNK_SIZE = 1024 ** 2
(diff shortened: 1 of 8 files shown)


pyperformance/data-files/benchmarks/bm_asyncio_websockets/run_benchmark.py
```diff
@@ -23,13 +24,20 @@ async def handler(websocket) -> None:
     stop.set()
 
 
+async def send(ws):
+    try:
+        await ws.send(DATA)
+    except websockets.exceptions.ConnectionClosedOK:
```
kumaraditya303, line 30: Does it happen on windows only or other platforms too?

kumaraditya303:
There's a failing CI of pypy which seems related to this

mdboom:
> There's a failing CI of pypy which seems related to this

Yeah -- I don't think there is a version of dask that is new enough to be compatible with Windows but old enough to work with PyPy.  I'm trying to figure out how to mark the benchmark as CPython-only, but I'm not sure there is a way.

pyperformance/data-files/benchmarks/bm_asyncio_websockets/run_benchmark.py
mdboom: Windows only.

kumaraditya303 (approved):
LGTM

PR #367, opened 2024-11-02
Add `argparse` benchmarks for subparsers and many optionals

savannahostrowski:
This PR adds two benchmarks for `argparse`. The first is for an argparse program with multiple subparsers, each with their own subcommands. The second is for an argparse program with many optional arguments.


mdboom (approved):
LGTM.  Thanks!

PR #369, opened 2024-11-20
Fix function to get if pyperformance is installed in editable mode

diegorusso:
This is used in `compile.py`: `compile_all` installs pyperformance for every version it compiles. If the main pyperformance is installed in editable mode, this should be propagated down.
The mechamism to do that was broken because was checking the presence of `setup.py`, now long gone.

### DIFF
--- a/pyperformance/__init__.py
+++ b/pyperformance/__init__.py
@@ -20,7 +20,7 @@ def is_installed():
 
 def is_dev():
     parent = os.path.dirname(PKG_ROOT)
-    return os.path.exists(os.path.join(parent, 'setup.py'))
+    return os.path.exists(os.path.join(parent, 'pyproject.toml'))
 
 
 def _is_venv():


mdboom (approved):
LGTM, but someone who has used it in this mode more than I have should probably also weigh in ...  (@ericsnowcurrently, @vstinner...?)

ericsnowcurrently:
I don't recall enough about this to be useful.  I'm sure Diego's fix is okay.

ericsnowcurrently (approved):
LGTM

diegorusso:
> I don't recall enough about this to be useful. I'm sure Diego's fix is okay.

Yes, this is one of the thing that is useful only if you use a specific feature of the tool. @mdboom keeps saying that they don't use the `compile_all` at all, maybe it's time for me to move away from it.

Issue #319, opened 2023-10-19
Editable installation isn't working properly


diegorusso:
If pyperformance is installed as editable `pip install -e .` from the git repository, it doesn't work when invoked

```
$ /tmp/tmpdir/prefix/bin/python3 -u -m pyperformance                                        
ERROR: pyperformance should not be run without installing first                                                                         
(consider using the dev.py script)             
```

The check is done by[ is_installed()](https://github.com/python/pyperformance/blob/main/pyperformance/__init__.py#L13C5-L13C17) which calls both `_is_dev()` and `_is_devel_install()` which are both broken.

The function [is_dev()](https://github.com/python/pyperformance/blob/main/pyperformance/__init__.py#L21) bases the logic on the presence of `setup.py`  which is long gone.

The function [_is_devel_install()](https://github.com/python/pyperformance/blob/main/pyperformance/__init__.py#L32) instead check the presence of `pyperformance.egg-link` in the site-packages.

In my installation I have:
* `pyperformance.egg-info` directory in the git repository
*  `__editable___pyperformance_1_0_9_finder.py`, `__editable__.pyperformance-1.0.9.pth`, `pyperformance-1.0.9.dist-info` in the site-packages

I'm trying to invoke pyperformance using a dev instance via 

```
python3.12 /work/oss/repos/pyperformance/dev.py compile_all benchmark.conf
```
and it will fail eventually:

```
2023-10-19 15:25:52,049: Building wheels for collected packages: pyperformance                                                                                                                                                                                                   
2023-10-19 15:25:52,050:   Building editable for pyperformance (pyproject.toml): started                                                                                                                                                                                         
2023-10-19 15:25:52,885:   Building editable for pyperformance (pyproject.toml): finished with status 'done'                                                                                                                                                                     
2023-10-19 15:25:52,888:   Created wheel for pyperformance: filename=pyperformance-1.0.9-0.editable-py3-none-any.whl size=6710 sha256=159fbc3f13ae47399e04d5f7cdf5e024c3652b2ed91096f8638844238e005c0d                                                                           
2023-10-19 15:25:52,888:   Stored in directory: /tmp/pip-ephem-wheel-cache-ubyuatck/wheels/0d/04/cb/add3bede8e376e8789c834f334f1834467606bde2625d5a4ad                                                                                                                           
2023-10-19 15:25:52,896: Successfully built pyperformance                                                                                                                                                                                                                        
[...]

PR #371, opened 2024-11-26
Fix _is_dev_install by using importlib.metadata API

diegorusso:
Instead of checking the presence of files in the filesystem, use the importlib.metadata API to check if the installation is in editable mode.
Fix related to https://github.com/python/pyperformance/issues/319

### DIFF
--- a/pyperformance/__init__.py
+++ b/pyperformance/__init__.py
@@ -1,5 +1,7 @@
+import json
 import os.path
 import sys
+from importlib.metadata import distribution
 
 
 VERSION = (1, 11, 0)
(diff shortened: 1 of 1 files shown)


mdboom (approved):
LGTM

FFY00 (approved):

FFY00:
This relies on the assumption the build system implements [PEP 610](https://peps.python.org/pep-0610/), but that's reasonable.

diegorusso:
@FFY00 that's correct! I thought using a more standard way to do it. Anyway this check is very specific to pyperformance and it is not generic for a random package, so it should be pretty safe to use it.

PR #376, opened 2025-01-18
.github: Add ARM64 CI

corona10:
See: https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/ and https://github.com/psf/pyperf/pull/212

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         # Test all supported versions on Ubuntu:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, ubuntu-24.04-arm]
         python: ["3.9", "3.10", "3.11", "3.12", "3.13"]
         experimental: [false]
         build: ['']
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,14 +37,7 @@ jobs:
           - os: ubuntu-latest
             python: "3.14-dev"
             experimental: true
-          - os: ubuntu-24.04-arm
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         # Test all supported versions on Ubuntu:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, ubuntu-24.04-arm]
         python: ["3.9", "3.10", "3.11", "3.12", "3.13"]
         experimental: [false]
         build: ['']
```
diegorusso, line 26: Do we want to include some experimental tasks on arm as well?

corona10: yeah we should add!

corona10:
@diegorusso 
https://github.com/python/pyperformance/actions/runs/12869515091/job/35878495572?pr=376

Looks like no build for 3.14-dev at this moment

diegorusso:
> @diegorusso https://github.com/python/pyperformance/actions/runs/12869515091/job/35878495572?pr=376
> 
> Looks like there is no build for 3.14-dev at this moment.

Nightly builds seem to support only amd64: https://launchpad.net/~deadsnakes/+archive/ubuntu/nightly

@asottile is there any plan to expand nightly builds to Arm platforms?

asottile:
I would have to look at capacity and it would add quite a bit of maintenance but it might not be that bad

diegorusso:
> I would have to look at capacity and it would add quite a bit of maintenance but it might not be that bad

Let me know if you need any help.

PR #378, opened 2025-01-23
Fix sqlglot normalize benchmark

AmirHesamSH:
Conjunction string is changing in every iteration. Now, conjunction string is created for every iteration.


AmirHesamSH:
The function "normalize.normalize" change the content of "conjunction" string to its normalized form. So, from the second iteration, "conjunction" is the normalized form of the original string. 
To fix this, I just set the input string for every iteration.

mdboom:
Thanks for the explanation.  That makes sense.

Since this will change the behavior of the benchmark, it will no longer be comparable to previously collected data from this benchmark.  In these situations, what we normally do is create a new benchmark -- e.g. `sqlglot2` -- with the fix.  Would you mind doing that?

PR #380, opened 2025-01-28
Fix sqlglot normalize benchmark

AmirHesamSH:
Update sqlglot to V2
Create a new conjunction string for every iteration. 
In the previous version, conjunction string would be modified to normalized version after the first iteration.


AmirHesamSH:
So, the pull request seems ready :)
Please, let me know if I need to modify anything.
I also updated the description.

mdboom (approved):
Thanks, this is really helpful.

PR #381, opened 2025-01-30
Add the 'btree' benchmark.

nascheme:
This adds a new benchmark that's intended to be a better exercise for the cyclic garbage collector, compared with the existing benchmarks in the suite.  The intention is to simulate an application that processes a large number of Python objects in memory, with some relatively complex set of references linking them together.   The set of objects should a mixture of GC (container) and non-GC objects since that kind of mixture would be typical in many applications.   I initially had the `NUM_NODES` parameter set to 3,000,000.  However, in order to keep the running time down and reduce memory usage, I decreased it to 200,000.

I should note that this benchmark doesn't exercise the GC in terms of breaking reference cycles.  The `create_gc_cycles` is kind of a micro benchmark of that.  I'd like to add an additional benchmark that more closely simulates an application that creates a bunch of objects but only some of them are cyclic garbage.  That would be separate from this PR though.


pyperformance/data-files/benchmarks/bm_btree/run_benchmark.py
```diff
@@ -0,0 +1,474 @@
...
+    """
+    Instance attributes:
+      root: BNode
+    """
+
+    __slots__ = ['root']
+
+    def __init__(self, node_constructor=BNode):
+        assert issubclass(node_constructor, BNode)
+        self.root = node_constructor()
```
AA-Turner, line 249: Is changing `node_constructor` ever used?

```suggestion
    def __init__(self):
        self.root = BNode()
```

pyperformance/data-files/benchmarks/bm_btree/run_benchmark.py
```diff
@@ -0,0 +1,474 @@
...
+
+    def keys(self):
+        return list(self.iterkeys())
+
+    def values(self):
+        return list(self.itervalues())
+
+    def __iter__(self):
+        for key in self.iterkeys():
+            yield key
```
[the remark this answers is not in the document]
nascheme: I removed most of them since they are not used and I removed the inheritance.  Bit simpler that way, IMHO.

pyperformance/data-files/benchmarks/bm_btree/run_benchmark.py
nascheme: Done.

AA-Turner (approved):

pyperformance/data-files/benchmarks/bm_btree/run_benchmark.py
```diff
@@ -0,0 +1,418 @@
...
+            if matches:
+                # Case 2.
+                if is_big(node):
+                    # Case 2a.
+                    extreme = node.get_max_item()
+                    node.delete(extreme[0])
+                    self.items[p] = extreme
+                elif is_big(upper_sibling):
+                    # Case 2b.
+                    extreme = upper_sibling.get_min_item()
```
maurycy, line 168: Where is this method defined?

pyperformance/data-files/benchmarks/bm_btree/run_benchmark.py
```diff
@@ -0,0 +1,418 @@
...
+                    upper_sibling.delete(extreme[0])
+                    self.items[p] = extreme
+                else:
+                    # Case 2c.
+                    extreme = upper_sibling.get_min_item()
```
maurycy, line 173: Where is this method defined?

pyperformance/data-files/benchmarks/bm_btree/run_benchmark.py
```diff
@@ -0,0 +1,418 @@
...
+                raise KeyError(key)
+        else:
+            node = self.nodes[p]
+            lower_sibling = p > 0 and self.nodes[p - 1]
+            upper_sibling = p < len(self.nodes) - 1 and self.nodes[p + 1]
+            if matches:
+                # Case 2.
+                if is_big(node):
+                    # Case 2a.
+                    extreme = node.get_max_item()
```
maurycy, line 163: Where is this method defined?

pyperformance/data-files/benchmarks/bm_btree/run_benchmark.py
```diff
@@ -0,0 +1,418 @@
...
+        self.e = e
+        self.f = f
+
+
+def make_records(num_nodes):
+    rnd = random.Random(RANDOM_SEED)
+    for node_id in range(num_nodes):
+        a = node_id
+        b = f'node {node_id}'
+        c = rnd.randbytes(node_id % 100)
```
maurycy, line 305: Introduced in Python 3.9, while:

https://github.com/python/pyperformance/pull/381/files#diff-5ac73097bc69dcf4f2f194138de6c0445c44b59541706d887b5902e8e54641f3R3

PR #384, opened 2025-02-11
`compile`: read `git_remote` from the `[scm]` section

nelhage:
The example configuration (`doc/benchmark.conf.sample`) and the configuration in `test_commands.py` both expect it to be there.

Add a fallback in case anyone has existing configurations with it set under `[config]`.


StanFromIreland:
Hello, sorry, this is a duplicate of https://github.com/python/pyperformance/pull/397

StanFromIreland:
Oops, I got my tabs messed up.

Issue #223, opened 2022-07-07
How should we be using dependabot in this repo?


mdboom:
I noticed a few dependabot PRs in this repo, such as [this one](https://github.com/python/pyperformance/pull/220) that upgrade dependencies of specific benchmarks.  While in general, I think this is good practice, for a benchmark suite, I think we'd want to upgrade these dependencies as infrequently as possible to keep benchmarking results comparable with one another (and not have to always rerun baselines). Occasionally we are forced to upgrade, for example to get compatibility with a new version of CPython, but that should be deliberate.

(It's possible there is a security counterargument to be made, but I'm not a security expert and I don't know specifically whether that matters or not).

Would it make sense to update the dependabot config to only look at the top-level dependencies of `pyperformance` itself rather than the dependencies of specific benchmarks?

ericsnowcurrently:
> for a benchmark suite, I think we'd want to upgrade these dependencies as infrequently as possible to keep benchmarking results comparable with one another

+1

> It's possible there is a security counterargument to be made

I don't consider the benchmarks to be a security concern, though I haven't done a thorough analysis of all the benchmark code.  IIRC, none of the benchmarks access anything other than files they already own or temporary network services they start themselves.

> Would it make sense to update the dependabot config to only look at the top-level dependencies of pyperformance itself

+1

mdboom:
It looks like this may not be easily possible: https://github.com/dependabot/dependabot-core/issues/4364

maurycy:
@mdboom @hugovk 

I believe this can be closed now, since #407 is merged.

AA-Turner:
Seems this hasn't worked, #412 was just opened. cc @maurycy.

maurycy:
@hugovk @AA-Turner 

The fix should be deployed:

- https://github.com/dependabot/dependabot-core/issues/4364#issuecomment-3289995774

Let's see how it goes before closing this issue.

🤞

maurycy:
@hugovk @AA-Turner It seems to be that we're good now?

hugovk:
We got a new one: https://github.com/python/pyperformance/pull/439

But this is out-of-schedule to fix a potential security issue, so I expect the config doesn't control this. I'll just close that one.

maurycy:
@hugovk Does #449 imply this is still a problem? 🫢

hugovk:
I think it was a security update and they're are exempt from the Dependabot config, so should be fine.

PR #391, opened 2025-05-21
Try skipping dependenabot updates on benchmarks

emmatyping:
Wanted to see if this might skip the benchmark dependencies which we want to keep stable.

See https://github.com/python/pyperformance/issues/223 for context.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,14 @@ updates:
     directory: "/pyperformance/requirements"
     schedule:
       interval: "monthly"
+  # Don't update test dependencies for reproducibility.
+  # Dependabot doesn't support excluding directories
+  # (https://github.com/dependabot/dependabot-core/issues/4364)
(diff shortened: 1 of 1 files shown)


hugovk:
I'm not sure if this config file applies to the Dependabot security related PRs?

Anyway, I pushed this to my fork's `main` and it gives an error:

> ### Your .github/dependabot.yml contained invalid details
> Dependabot encountered the following error when parsing your .github/dependabot.yml:
> 
> `The property '#/updates/1/directories' of type string did not match the following type: array`
> 
> Please update the config file to conform with Dependabot's specification.
> 
> For more info on the config file format, see [the config file documentation](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#configuration-options-for-dependabotyml)

https://github.com/hugovk/pyperformance/runs/42703898168

emmatyping:
I fixed the syntax on my fork but it still didn't work :(

PR #397, opened 2025-06-07
Fix git_remote option in compile.py

zhaozhixu:
`git_remote` option is one of "scm" options instead of "config" options.


StanFromIreland:
Hello, sorry, this is a duplicate of https://github.com/python/pyperformance/pull/384.

PR #398, opened 2025-06-09
Add a comparison output style for GitHub tables

ZeroIntensity:
After running benchmarks for a PR of mine, I noticed that `--output_style table` formats it in a way that GitHub doesn't know how to deal with. Instead of manually making one, I thought it would be much more convenient if there were a way to emit a [GitHub-style](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-tables) table that could be directly copy-pasted to a comment. I guess this is technically a markdown table, but there's too many different parsers for markdown to conform to that standard.

I didn't bother with adding padding or anything, so it looks like garbage in the terminal, but it looks fine on GitHub.

### DIFF
--- a/doc/usage.rst
+++ b/doc/usage.rst
@@ -185,7 +185,7 @@ options::
   -v, --verbose         Print more output
   -O STYLE, --output_style STYLE
                         What style the benchmark output should take.
-                        Valid options are 'normal' and 'table'.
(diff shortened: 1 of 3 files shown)


ZeroIntensity:
Eh, I'm having second thoughts about this one.

PR #401, opened 2025-06-20
Pass --warmup along to pyperf

mdboom:
When running pystats collection, it is useful to /not/ do warmup runs so we don't "miss" any of the stats.  This requires being able pass the `--warmups` argument to the underlying pyperf.

### DIFF
--- a/pyperformance/run.py
+++ b/pyperformance/run.py
@@ -247,6 +247,7 @@ def get_pyperf_opts(options):
     if options.hook:
         for hook in options.hook:
             opts.append('--hook=%s' % hook)
+    # --warmups=0 is a valid option, so check for `not None` here
     if options.warmups is not None:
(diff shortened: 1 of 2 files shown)


pyperformance/run.py
```diff
@@ -247,5 +247,7 @@ def get_pyperf_opts(options):
     if options.hook:
         for hook in options.hook:
             opts.append('--hook=%s' % hook)
+    if options.warmups is not None:
```
corona10, line 250:
```suggestion
    if options.warmups:
```

corona10 (changes requested):
Leave the nit change.

pyperformance/run.py
corona10: Ah got it

corona10: Can you leave the comment for it?

PR #409, opened 2025-09-05
Use a relative `exclude-paths` in `dependabot.yml` and a wildcard

maurycy:
Unfortunately, #407 is incorrect, as demonstrated by #408: `/pyperformance/data-files/benchmarks/bm_dulwich_log` should be exluded.

It's not exactly clear to me why:

1. is `exclude-paths` relative to the `directory`?
2. is it because `exclude-paths` needs a wildcard?
3. is it because `exclude-paths` doesn't care about security updates?

It's impossible to test, and there are three sources of knowledge on the subject:

- The docs: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference
- The thread, where Dependabot developers reply: https://github.com/dependabot/dependabot-core/issues/4364
- Existing `dependabot.yaml` files across the Github universe: https://github.com/search?q=path%3Adependabot.yml+exclude-paths&type=code

My – possibly incorrect – understanding is that 1. is true, 2. is likely true, and 3. is not definite (does "update PR creation" include security?) from reading comments:

https://github.com/dependabot/dependabot-core/issues/4364#issuecomment-3250249710

I think that we can revert #296 once we confirm that it's _indeed_ the fix.


hugovk:
🤞

.github/dependabot.yml
```diff
@@ -1,8 +1,8 @@
 version: 2
 updates:
   - package-ecosystem: "pip"
-    directory: "/pyperformance/requirements"
+    directory: "/"
     exclude-paths:
-      - "pyperformance/data-files/"
+      - "pyperformance/data-files/**"
```
dnlfm, line 6: Interesting.. The previous version should have worked as well. May be some regression in how Dependabot is trying to filter paths @hugovk

hugovk: Or maybe this is working?

https://github.com/python/pyperformance/pull/411 is an expected update in the top-level pyproject.toml

PR #405, opened 2025-09-01
Add pre-commit, `ruff check` configuration and Github action

maurycy:
After #404, I ran `uvx ruff check` and found a lot of not so subtle errors:

```
12:33:31.816619000AM CEST maurycy@gimel /Users/maurycy/src/pyperformance % uvx ruff check --output-format concise
pyperformance/_benchmark.py:4:5: F822 Undefined name `Benchmarkcheck_name` in `__all__`
pyperformance/_benchmark.py:221:67: F821 Undefined name `NOOP`
pyperformance/_benchmark_metadata.py:197:13: F841 Local variable `repo` is assigned to but never used
pyperformance/_manifest.py:346:17: F821 Undefined name `seclines`
pyperformance/_manifest.py:366:9: F841 Local variable `yielded` is assigned to but never used
pyperformance/_pyproject_toml.py:103:15: F821 Undefined name `ValuError`
pyperformance/_pyproject_toml.py:125:30: F541 [*] f-string without any placeholders
pyperformance/_pyproject_toml.py:129:9: F841 Local variable `text` is assigned to but never used
pyperformance/_utils.py:9:5: F822 Undefined name `run_command` in `__all__`
pyperformance/data-files/benchmarks/bm_2to3/run_benchmark.py:20:16: F401 `lib2to3` imported but unused; consider using `importlib.util.find_spec` to test for availability
pyperformance/data-files/benchmarks/bm_2to3/vendor/src/lib2to3/btm_matcher.py:162:16: E721 Use `is` and `is not` for type comparisons, or `isinstance()` for isinstance checks
pyperformance/data-files/benchmarks/bm_2to3/vendor/src/lib2to3/btm_utils.py:170:17: F841 Local variable `leaf` is assigned to but never used
pyperformance/data-files/benchmarks/bm_2to3/vendor/src/lib2to3/fixes/fix_except.py:48:9: F841 Local variable `syms` is assigned to but never used
pyperformance/data-files/benchmarks/bm_2to3/vendor/src/lib2to3/fixes/fix_exec.py:28:9: F841 Local variable `syms` is assigned to but never used
pyperformance/data-files/benchmarks/bm_2to3/vendor/src/lib2to3/fixes/fix_exitfunc.py:66:13: F841 Local variable `stmt_container` is assigned to but never used
pyperformance/data-files/benchmarks/bm_2to3/vendor/src/lib2to3/fixes/fix_has_key.py:80:9: F841 Local variable `anchor` is assigned to but never used
[...]

PR #410, opened 2025-09-05
Add `ruff format` to pre-commit, run it, except the benchmarks

maurycy:
This should be the last one from this series:

https://github.com/python/pyperformance/pull/405#issuecomment-3245226963


.github/workflows/pre-commit.yml
```diff
@@ -1,12 +1,12 @@
-name: Lint
+name: pre-commit
```
maurycy, line 1: It makes more sense to me, but I'm OK with renaming back, since `lint.yml` is the pattern:

https://github.com/search?q=org%3Apython%20action-pre-commit-uv&type=code

PR #413, opened 2025-09-13
Incorrect `isinstance(readme, 'str')` check

maurycy:
`'str'` is incorrect there:

```
>>> isinstance('meow', 'str')
Traceback (most recent call last):
  File "<python-input-0>", line 1, in <module>
    isinstance('meow', 'str')
    ~~~~~~~~~~^^^^^^^^^^^^^^^
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
>>> isinstance('meow', str)
True
```

### DIFF
--- a/pyperformance/_pyproject_toml.py
+++ b/pyperformance/_pyproject_toml.py
@@ -167,7 +167,7 @@ def _normalize_project(data, rootdir, name, requirefiles, **_ignored):
     key = 'readme'
     if key in data:
         readme = data[key]
-        if isinstance(readme, 'str'):
+        if isinstance(readme, str):
             readme = data[key] = {'file': readme}
         # XXX Check the suffix.
         # XXX Handle 'content-type'.


maurycy:
If I find more similarly obvious issues, I will create one PR for them.

maurycy:
@hugovk 

> Looks like we're missing tests for this bit of code. Would it be possible to add something?

Definitely! Thank you for spotting this.

What do you think about bd5e2d208a7a2590a9bca0dec23c3c372561be13? It fails on the `main` branch.

Personally, I'm not a huge fan of this, as it's importing an `_internal` module. It seems to be the rule, though:

- https://github.com/python/pyperformance/blob/main/pyperformance/tests/test_python.py#L4
- https://github.com/python/pyperformance/blob/main/pyperformance/tests/test_pythoninfo.py#L7
- https://github.com/python/pyperformance/blob/main/pyperformance/tests/test_venv.py#L8

If you like this stub, I can extend it with more tests.

I have to admit that #109 wasn't a particularly lucky PR:

https://github.com/python/pyperformance/pull/109/files#diff-3e8f1cf61bbb100da8f1596584dbbfbdab1381669a6c4a036246bd19b6eea3c7R191-R193

https://github.com/python/pyperformance/pull/109/files#diff-3e8f1cf61bbb100da8f1596584dbbfbdab1381669a6c4a036246bd19b6eea3c7R248-R251

https://github.com/python/pyperformance/pull/109/files#diff-37e368652aee97ec2843ef7927faa34e8e5d681315643af43890e02b8448868bR31

https://github.com/python/pyperformance/pull/109/files#diff-71569863b6c2ec4af655ac7c8a3a9af1e12f7a4ee2d6c571d74f696c880949c8R115

https://github.com/python/pyperformance/pull/109/files#diff-3e8f1cf61bbb100da8f1596584dbbfbdab1381669a6c4a036246bd19b6eea3c7R29-R36

More generally, one could wonder if the project should reimplement PEP 621.

pyperformance/tests/test_pyproject_toml.py
```diff
@@ -0,0 +1,98 @@
...
+import tempfile
+import pathlib
+from pyperformance import _pyproject_toml
+
+
+class TestPyProjectToml(unittest.TestCase):
+    def test_parse_project_with_readme_string(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            tmp_path = pathlib.Path(tmpdir)
+            toml_content = """
```
maurycy, line 11: https://docs.python.org/3/library/textwrap.html#textwrap.dedent

hugovk (approved):
Thanks!

PR #414, opened 2025-09-15
Drop support for Python 3.9, test PyPy3.11 instead of PyPy3.10

hugovk:
It's almost end-of-life:

* https://devguide.python.org/versions/
* https://peps.python.org/pep-0596/

Also move CI from PyPy3.10 to PyPy3.11 because PyPy only includes PyPy3.11 in the latest releases: https://pypy.org/posts/2025/07/pypy-v7320-release.html

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,7 +43,7 @@ jobs:
             build: 'free-threading'
           # Also test PyPy, macOS, and Windows:
           - os: ubuntu-latest
-            python: pypy-3.10
+            python: pypy-3.11
             experimental: false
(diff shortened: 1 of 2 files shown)


maurycy (approved):
@hugovk 

What do you think about updating `requires-python` in `data-files`, too?

- https://github.com/search?q=repo%3Apython%2Fpyperformance+%22requires-python%22+%223.8%22&type=code
- https://github.com/search?q=repo%3Apython%2Fpyperformance%20%22Programming%20Language%20%3A%3A%20Python%20%3A%3A%203.7%22&type=code

PR #417, opened 2025-10-23
Bump a few dependencies to make it work with main (3.15)

diegorusso:


hugovk:
Updated from `main` to include https://github.com/python/pyperformance/pull/421.

hugovk (approved):
macOS has been failing for 3.12, known issue: https://github.com/python/pyperformance/issues/422
Also marked experimental, so doesn't fail the build.

3.15 free-threaded is because psutil doesn't have any free-threaded wheels yet. They'll have 3.13 soon, and hopefully 3.14 soonish after that, so perhaps we can test 3.14 free-threaded. But we can leave it as experimental for now.

https://github.com/giampaolo/psutil/issues/2565#issuecomment-3439154482

PR #420, opened 2025-10-24
Prepare release 1.12.0

diegorusso:


doc/changelog.rst
```diff
@@ -1,10 +1,33 @@
...
 * Bump dask[distributed] to 2024.10.1 for Windows compatibility
 * Bump greenlet to 3.1.0 for compatibility with 3.13
 * Bump tornado to 6.2.0
+* Bump pyperf to 2.9.0
+* Bump sqlglot to V2
+* Bump mypy to 1.18.2
+* Fix check for editable mode
+* Fix --same-loops handling
+* Fix: pass --warmup and --timeout flags to pyperf
+* Add xdsl benchmark
```
hugovk:
```suggestion
```
If including https://github.com/python/pyperformance/pull/421

diegorusso: I would say to add it but then say that is disabled
```
* Add xdsl benchmark (but disabled by default due to a regression)
```
WDYT?

PR #423, opened 2025-10-24
Remove redundant code for EOL Python 3.7 and older

hugovk:
We have `requires-python = ">=3.10"` in `pyproject.toml` to ensure we don't install on an EOL version.


corona10 (approved):
lgtm, Thanks for the clean up

Issue #416, opened 2025-10-08
Over 20 benchmarks fail to run on Python 3.14.0


Lewoniewski:
After updating to Python 3.14.0, more than 20 benchmarks fail to run that did not occur in previous versions (e.g., Python 3.13.x). This issue was reproduced on Win 11 with AMD and Intel processors. 

Will this issue be addressed or fixed for compatibility with Python 3.14?

hugovk:
https://github.com/python/pyperformance/pull/417 was merged earlier, thanks for testing!

maurycy:
All seem good:

- [typescript.3.10.19.txt](https://github.com/user-attachments/files/23135908/typescript.3.10.19.txt)
- [typescript.3.12.12.txt](https://github.com/user-attachments/files/23135910/typescript.3.12.12.txt)
- [typescript.3.13.9.txt](https://github.com/user-attachments/files/23135909/typescript.3.13.9.txt)

PR #425, opened 2025-10-24
Use `uv venv` and `uv pip`

maurycy:
This is a conversation starter now: should we gradually start moving towards `uv` that handles custom builds, venvs and pinned runs more efficiently?

The code already works, confirmed with:

```
python dev.py run
```

Tests are 2x faster:

- [last main run](https://github.com/python/pyperformance/actions/runs/18779843038) took 13m57s
- [3a402551d306ba7b4e80c1b308c426f5f2a08e30](https://github.com/python/pyperformance/actions/runs/18808706662), with [enable-cache](https://docs.astral.sh/uv/guides/integration/github/#caching) in [the workflow](https://github.com/python/pyperformance/pull/425/files#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R71), took 7m14s
- [4b5f47a1292e94802af93873709b688cc0c43de1](https://github.com/maurycy/pyperformance/actions/runs/18808881710), without caching in the action, took 8m20s

I didn't want to do too many changes at once, but in my opinion the best direction would be iterating over each benchmark with `uv`. `pyperformance` is just a benchmark suite after all, and the whole venv, custom build etc. management makes it unnecessarily complex.

Notably, `uv run` has [--project](https://docs.astral.sh/uv/reference/cli/#uv-tool-install--project), [--isolated](https://docs.astral.sh/uv/reference/cli/#uv-tool-run--isolated) (now it's isolated [if the requirements cannot be met](https://github.com/python/pyperformance/blob/74a5b03b4b97b51ae793d19691933852177d579f/pyperformance/run.py#L103-L123)), [--with-requirements](https://docs.astral.sh/uv/reference/cli/#uv-run--with-requirements) and [--python](https://docs.astral.sh/uv/reference/cli/#uv-tool-run--python) (supporting custom builds), eg:

```
8:41:50.613123804PM CEST maurycy@gunnbjorn ~/pyperformance (uv) % uv run --isolated --project=benchmarks/bm_sympy --python=/home/maurycy/cpython/python benchmarks/bm_sympy/run_benchmark.py
Installed 4 packages in 12ms
.....................
sympy_expand: Mean +- std dev: 281 ms +- 1 ms
.....................
sympy_integrate: Mean +- std dev: 12.8 ms +- 0.1 ms
.....................
sympy_sum: Mean +- std dev: 92.8 ms +- 0.8 ms
.....................
sympy_str: Mean +- std dev: 161 ms +- 1 ms
```

This is similar to what happens now, just merging [temporary outputs](https://github.com/python/pyperformance/blob/74a5b03b4b97b51ae793d19691933852177d579f/pyperformance/_benchmark.py#L223-L228).

From what I measured venv management and dependencies only take ~5% of the `run` time, so no dramatic speed up should be expected, but the simplification might be worth it. Important to keep in mind that many features are never ever used, if [a large chunk of benchmarks stays broken for so long](https://github.com/python/pyperformance/issues/416#issue-3496543536) and [some obvious](https://github.com/python/pyperformance/commit/354b293047566d10bd28f12188846200c44143b6#diff-3e8f1cf61bbb100da8f1596584dbbfbdab1381669a6c4a036246bd19b6eea3c7L178) issues lurked for a while.

Another elephant in the room is [compile](https://github.com/python/pyperformance/blob/74a5b03b4b97b51ae793d19691933852177d579f/pyperformance/compile.py#L279) that could also be outsourced.

I'm more than open to any suggestions!


corona10:
Unless you have a strong motivation to maintain this project based on uv, let's keep it as is based on standard(pip).
I love uv but this should consider other factors.

PR #438, opened 2025-11-03
Add additional gc benchmark with pickletools (#437)

### LINKED ISSUE #437: Add additional GC related benchmark
Given this [performance regression in Python 3.14](https://github.com/python/cpython/issues/139951), it would be nice if we had a benchmark which would have more clearly shown it before final release.  I have a [new GC specific benchmark](https://github.com/python/pyperformance/pull/381) but it doesn't do things that would show this regression.

We should consider adding a new benchmark.  The key features would be:
- creating a bunch of net new container objects, in order to trigger many, potentially young generation, collections
- creating tuple objects that can be untracked by the GC, in order to avoid slowing down full collections

Ideally we prefer to avoid a micro or synthetic benchmark and instead find some kind of real application that shows this regression.   E.g. near the 3.13 release, we found a GC regression shown by Sphinx building Python docs.  Something like that would be good.

pgdr:
Adds a benchmark reproducing the Python 3.14 garbage collector regression described in [cpython/#140175](https://github.com/python/cpython/issues/140175).

This real-world case uses `pickletools` to demonstrate the performance issue.

Fixes #437.

### DIFF
--- a/pyperformance/data-files/benchmarks/bm_pickle_opt/run_benchmark.py
+++ b/pyperformance/data-files/benchmarks/bm_pickle_opt/run_benchmark.py
@@ -6,38 +6,23 @@ Python 3.14.0 had a performance regression, see
 
 """
 
-import tempfile
-from pathlib import Path
-import pyperf
 import pickle
 import pickletools
+import pyperf
 
 
-def setup(fname, N):
-    x = {}
-    for i in range(1, N):
-        x[i] = f"ii{i:>07}"
-
-    with open(fname, "wb") as fh:
-        pickle.dump(x, fh, protocol=4)
+def setup(N: int) -> bytes:
+    x = {i: f"ii{i:>07}" for i in range(N)}
+    return pickle.dumps(x, protocol=4)
 
 
-def run(fname):
-    with open(fname, "rb") as fh:
-        p = fh.read()
-
-    s = pickletools.optimize(p)
-
-    with open(fname.with_suffix(".out"), "wb") as fh:
-        fh.write(s)
+def run(p: bytes) -> None:
+    pickletools.optimize(p)
 
 
 if __name__ == "__main__":
     runner = pyperf.Runner()
-    N = 1_000_000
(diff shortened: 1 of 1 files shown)


pgdr:
@sergey-miryanov Thanks for the review.  I have fixed all issues you pointed out.

pgdr:
@sergey-miryanov Something strange happens here.  Even though I use the context manager (`tempfile.TemporaryDirectory`), occasionally when I kill pyperformance, the directory remains not cleaned up.

I am not able to reproduce this behavior when not running with `pyperf`, though, so it might be related to the way `pyperf` sets up (parallel?) runners.

It sounds like a bug, but I can't tell where.

sergey-miryanov (approved):
Code looks good to me.

nascheme:
> These tests (and the PR) has N = 1'000'000. The downside is that running the benchmark (with Python 3.14) takes almost 10 minutes.

Taking 10 minutes would be too long.  However, it only takes about 6 seconds for me to run this on Python 3.14.0, on my hardware.  Perhaps the 10 minutes is for when N = 10e6?  The regression I see from 3.13 to 3.14 with N = 1e6 seems large enough (1.5 seconds vs 6 seconds, roughly).

Nice work on this benchmark.  I think it's good because `optimize()` is doing some meaningful work, unlike some other synthetic benchmarks.  In addition to showing this regression in the GC, I would expect this benchmark to catch other kinds of performance regressions.

nascheme:
Small suggestion: it would be simpler to use `io.BytesIO()` rather than using real files in a temporary folder.  I don't think that affects the usefulness of the benchmark, since we are not really testing real file IO speed.  Something like this:
```
def setup(fp, N):
    x = {}
    for i in range(1, N):
        x[i] = f"ii{i:>07}"
    pickle.dump(x, fp, protocol=4)

def run(fp):
    p = fp.read()
    s = pickletools.optimize(p)
```
You could use `dumps()` as well and do away with the file.

PR #447, opened 2025-12-29
Add bm_base64 covering common base64 module APIs

gpshead:
Motivation https://github.com/python/cpython/pull/143262 but also in general this wasn't covered and @serhiy-storchaka is also doing work in this area that'll become relevant such as https://github.com/python/cpython/pull/143216.


gpshead:
(draft while i deal with more plumbing test failures)

pyperformance/data-files/benchmarks/bm_base64/run_benchmark.py
```diff
@@ -0,0 +1,274 @@
...
+"""
+
+import base64
+import random
+import pyperf
+
+
+# Generate test data with fixed seed for reproducibility
+random.seed(12345)
+DATA_TINY = bytes(random.randrange(256) for _ in range(20))
```
serhiy-storchaka: `randbytes()` is usually much faster and memory efficient. Or `os.urandom()`.

pyperformance/data-files/benchmarks/bm_base64/run_benchmark.py
```diff
@@ -0,0 +1,274 @@
...
+# --- Base64 (includes validate=True) ---
+
+def bench_b64_small(loops):
+    range_it = range(loops)
+    t0 = pyperf.perf_counter()
+    for _ in range_it:
+        for _ in range(450):
+            base64.b64encode(DATA_TINY)
+            base64.b64decode(B64_TINY)
+            base64.b64decode(B64_TINY, validate=True)
```
serhiy-storchaka: Decoding has twice larger weight than encoding. `validate=True` should not affect performance for such input in reasonable implementation.

On other hand, I would test decoding from an ASCII string.

If you keep several decoding calls, you should balance them by equal number of encoding calls.

pyperformance/data-files/benchmarks/bm_base64/run_benchmark.py
```diff
@@ -0,0 +1,274 @@
...
+            base64.b64decode(B64_LARGE, validate=True)
+        base64.b64encode(DATA_HUGE)
+        base64.b64decode(B64_HUGE)
+        base64.b64decode(B64_HUGE, validate=True)
+    return pyperf.perf_counter() - t0
+
+
+# --- URL-safe Base64 (small only) ---
+
+def bench_urlsafe_b64_small(loops):
```
serhiy-storchaka, line 100: They may be more rare, but there are more than 2 standard Base64 variants. For example, the variant with `altchars='+,'` is used for IMAP mailbox names (RFC 3501). It may be not worth to add benchmarks for them, but there would be a difference. `urlsafe_*()` are more optimized than `altchars=`.

pyperformance/data-files/benchmarks/bm_base64/run_benchmark.py
```diff
@@ -0,0 +1,274 @@
...
+            base64.urlsafe_b64encode(DATA_MEDIUM)
+            base64.urlsafe_b64decode(B64_URLSAFE_MEDIUM)
+        base64.urlsafe_b64encode(DATA_9K)
+        base64.urlsafe_b64decode(B64_URLSAFE_9K)
+    return pyperf.perf_counter() - t0
+
+
+# --- Base32 ---
+
+def bench_b32_small(loops):
```
serhiy-storchaka, line 120: There are also several variants. Some of them are implemented with preprocessing or postprocessing, this can affect performance.

serhiy-storchaka (commented):
What are the results and the total time?

My suggestions:

* Test an ASCII string input for decoding.
* Balance encoding and decoding.

pyperformance/data-files/benchmarks/bm_base64/run_benchmark.py
gpshead: In general the variants are just a str.translate or equivalent post processing step so I didn't think there was a lot of value in covering those for performance.  a85 and b85 seemed different enough, and wrapcol is arguably "just" a memory move benchmark but covering that for the API that already had it seemed more meaningful as that could be implemented a variety of ways.

pyperformance/data-files/benchmarks/bm_base64/run_benchmark.py
gpshead: eyeballing them I don't know if they're really worth covering.

pyperformance/data-files/benchmarks/bm_base64/run_benchmark.py
gpshead: I always forget that exists. thanks!

gpshead:
Example results comparing before and after https://github.com/python/cpython/pull/143262.

```
main-before-opt-b64.json
========================

Performance version: 1.14.0
Python version: 3.15.0a3+ (64-bit) revision 6b9a6c6ec3b
Report on Linux-6.17.0-8-generic-x86_64-with-glibc2.42
Number of logical CPUs: 15
Start date: 2026-01-02 06:47:53.863590
End date: 2026-01-02 06:54:17.985677

main-after-opt-b64.json
=======================

Performance version: 1.14.0
Python version: 3.15.0a3+ (64-bit) revision 61fc72a4a43
Report on Linux-6.17.0-8-generic-x86_64-with-glibc2.42
Number of logical CPUs: 15
Start date: 2026-01-02 06:30:08.791666
End date: 2026-01-02 06:36:26.047033

### ascii85_large ###
Mean +- std dev: 1.04 sec +- 0.01 sec -> 1.03 sec +- 0.01 sec: 1.01x faster
Not significant

### ascii85_small ###
Mean +- std dev: 20.7 ms +- 0.3 ms -> 20.6 ms +- 0.4 ms: 1.00x faster
Not significant

### base16_large ###
Mean +- std dev: 6.39 ms +- 0.06 ms -> 6.36 ms +- 0.04 ms: 1.00x faster
Not significant

### base16_small ###
Mean +- std dev: 376 us +- 3 us -> 372 us +- 3 us: 1.01x faster
Not significant

### base32_large ###
Mean +- std dev: 585 ms +- 8 ms -> 586 ms +- 7 ms: 1.00x slower
Not significant

### base32_small ###
Mean +- std dev: 11.4 ms +- 0.1 ms -> 11.5 ms +- 0.1 ms: 1.00x slower
Not significant

### base64_large ###
Mean +- std dev: 4.66 ms +- 0.03 ms -> 2.06 ms +- 0.01 ms: 2.26x faster
Significant (t=749.12)

### base64_small ###
Mean +- std dev: 314 us +- 5 us -> 253 us +- 5 us: 1.24x faster
Significant (t=64.73)

### base85_large ###
Mean +- std dev: 432 ms +- 4 ms -> 431 ms +- 4 ms: 1.00x faster
Not significant

### base85_small ###
Mean +- std dev: 8.37 ms +- 0.08 ms -> 8.39 ms +- 0.12 ms: 1.00x slower
Not significant

### urlsafe_base64_small ###
Mean +- std dev: 511 us +- 5 us -> 445 us +- 4 us: 1.15x faster
Significant (t=80.03)
```

serhiy-storchaka (approved):
I am going to add also `ignorechars` to support decoding multi-line data. So it may be worth to separate "raw" and "multi-line" cases, if we want to compare the effect of optimization. We could also separate encoding and decoding. But then we will have too much tests. So this is up to you.

gpshead:
yeah i'd initially started out with separate encode/decode but the matrix of benchmark cases got large and pyperformance is not really intended as a microbenchmark.  improvements in any of these will be reflected in the overall scores.

PR #451, opened 2026-01-11
Let pyperformance decide unpack sequence loops

### LINKED ISSUE #450: unpack_sequence benchmark too fast on modern computers
On modern computers, unpack_sequence completes in nanoseconds (!!!).

This is basically within the range of noise for modern computers. The cause is that for some reason, the inner loops are hardcoded instead of letting pyperf decide

https://github.com/python/pyperformance/blob/main/pyperformance/data-files/benchmarks/bm_unpack_sequence/run_benchmark.py#L457

The fix is simple: just remove the inner_loops argument. However, fixing this would need a major version bump, as all past unpack sequence results would be invalidated.

Fidget-Spinner:
Fixes https://github.com/python/pyperformance/issues/450

### DIFF
--- a/pyperformance/data-files/benchmarks/bm_unpack_sequence/run_benchmark.py
+++ b/pyperformance/data-files/benchmarks/bm_unpack_sequence/run_benchmark.py
@@ -454,4 +454,4 @@ if __name__ == "__main__":
     else:
         func = bench_all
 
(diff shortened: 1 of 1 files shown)


corona10 (commented):
Change looks good to me. But I prefer to discuss this change at CHIPS because it will affect to benchmark history.

PR #453, opened 2026-01-19
Enable decimal benchmarks in the test suite (MANIFEST)

skirpichev:
They were not added here in #341 for no reasons.

### DIFF
--- a/pyperformance/data-files/benchmarks/MANIFEST
+++ b/pyperformance/data-files/benchmarks/MANIFEST
@@ -37,9 +37,9 @@ chaos	<local>
 comprehensions	<local>
 crypto_pyaes	<local>
 dask	<local>
-deepcopy	<local>
 decimal_factorial	<local>
 decimal_pi	<local>
+deepcopy	<local>
 deltablue	<local>
 django_template	<local>
 dulwich_log	<local>
--- a/pyperformance/data-files/benchmarks/MANIFEST
+++ b/pyperformance/data-files/benchmarks/MANIFEST
@@ -37,8 +37,7 @@ chaos	<local>
 comprehensions	<local>
 crypto_pyaes	<local>
 dask	<local>
-# FIXME: this doesn't work with `_pydecimal`
-# decimal_factorial	<local>
+decimal_factorial	<local>
 decimal_pi	<local>
 deepcopy	<local>
 deltablue	<local>
--- a/pyperformance/data-files/benchmarks/MANIFEST
+++ b/pyperformance/data-files/benchmarks/MANIFEST
@@ -37,7 +37,8 @@ chaos	<local>
 comprehensions	<local>
 crypto_pyaes	<local>
 dask	<local>
-decimal_factorial	<local>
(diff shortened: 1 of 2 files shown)


pyperformance/data-files/benchmarks/MANIFEST
```diff
@@ -38,6 +38,8 @@ comprehensions	<local>
 crypto_pyaes	<local>
 dask	<local>
 deepcopy	<local>
+decimal_factorial	<local>
+decimal_pi	<local>
```
hugovk, line 41: a-z
```suggestion
decimal_factorial	<local>
decimal_pi	<local>
deepcopy	<local>
```

hugovk (approved):

hugovk:
cc @mdboom in case there was a reason to omit them.

skirpichev:
Ah, pypy3 failure seems relevant.  decimal_factorial looks too slow on PyPy3 (with only pure-Python version of the decimal module).

pyperformance/data-files/benchmarks/MANIFEST
```diff
@@ -37,6 +37,8 @@ chaos	<local>
 comprehensions	<local>
 crypto_pyaes	<local>
 dask	<local>
+decimal_factorial	<local>
```
skirpichev, line 40:
```suggestion
# FIXME: this doesn't work with `_pydecimal`
# decimal_factorial	<local>
```

skirpichev:
@hugovk, anything else?

I think we can enable decimal_factorial test, but this will need a workaround for pure-Python decimal module.  The presence of accelerated module we can detect, trying to import the ``_decimal`` module.

1. Either we can lower ``n`` and ``c.prec`` (we can't use MAX_PREC for pure-Python module) in ``bench_decimal_factorial()``
2. or use different settings conditionally, for pure-Python and C modules

skirpichev:
Ok, adapted version of the factorial benchmark works with pure-Python module.  Timings (from CI): 287 ms on CPython 3.14 vs 2.36 sec on PyPy.

Edit: *with* the patch for python/cpython#140036 I got following on decimal_factorial benchmark with pure-Python class:
```
$ ./python Modules/_decimal/tests/bench.py

# ======================================================================
#                               Factorial
# ======================================================================

n = 100000

cdecimal:
calculation time: 26.301405s
conversion time: 0.000021s

int:
calculation time: 0.809898s
conversion time: 0.441830s


n = 1000000

cdecimal:
calculation time: 397.983149s
conversion time: 0.000115s

int:
calculation time: 39.675428s
conversion time: 10.546071s
```
So, probably we will have to adjust numbers anyway.

skirpichev:
CC @vstinner

vstinner:
Sorry, I'm not working on pyperformance anymore. It's uneasy for me to judge if this benchmark would be a good addition and it would measure correctly the big integer/decimal peerformance.

skirpichev:
The problem is that there is no decimal benchmarks at all.

PR #461, opened 2026-03-31
Support PyPy `_pickle` accelerator in `bm_pickle`

mattip:
I am working on adding a _pickle RPython implementation to PyPy. I ran into a problem where the current benchmark will ignore that possibility and will always use the pure-python version on PyPy. This PR makes the logic work for the case that PyPy does have a _pickle module and, like CPython, it is activated by default, but retains backward compatibility.


StanFromIreland (approved):
LGTM

PR #467, opened 2026-04-24
Hash pin GitHub Actions

hugovk:
Yet another compromise via unpinned GitHub Actions: https://socket.dev/blog/bitwarden-cli-compromised

Let's hash-pin GHA for our PyPI packages.

Done via `uvx gha-update`.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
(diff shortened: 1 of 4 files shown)


maurycy:
@hugovk 

- Does it effectively disable Dependabot?
- What would be the process for rerunning `uvx gha-update`?

The same would apply to https://github.com/psf/pyperf/pull/237 I think

hugovk:
> * Does it effectively disable Dependabot?

No, Dependabot will still run and update to the next available SHA, and also update the version in the comment.

> * What would be the process for rerunning `uvx gha-update`?

Just leave it to Dependabot. The cooldown means it waits that long before an update is suggested. Security updates will still come through immediately without needing to wait.

maurycy (approved):

sethmlarson (approved):

PR #475, opened 2026-05-26
Fix --same-loops flag in compile subprocess invocation

stratakis:
When the compile subcommand spawns a "pyperformance run" subprocess, it constructs --same_loops on the command line where argparse only accepts --same-loops. Change the flag to use the hyphenated form.


vstinner:
Merged, thanks for the fix.

